Issue #8, opened 2016-02-16
Python 3 compatiblity


andor44:
I haven't had time to find all the Python 3 incompatibilities but there's one I just ran into [here](https://github.com/rochacbruno/dynaconf/blob/524fd963c9c0b9965d4323e83f2086357f84ec5e/dynaconf/base.py#L316). `basestring` is not a thing in Python 3 anymore.

rochacbruno:
Thanks @andor44 I will work on implementing all the possibilities using **six** package alternatives.

aholik:
A note on Redis backend:

Python 3 receives bytes from Redis, I had to use `'decode_responses': True` option in `REDIS_FOR_DYNACONF` otherwise the loader failed.

rochacbruno:
Yeah another problem with namespaces https://repl.it/EHWD

can someone take a look?

rochacbruno:
@aholik I should put a note on docs for that.

rochacbruno:
Covered in 0.3.0 version thanks you all

PR #9, opened 2016-02-19
Just preparing tests for multiple python versions

ellisonleao:
Also:
- fix docs path on gitignore
- adding tox on test.req

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,8 +50,8 @@ coverage.xml
 # Django stuff:
 *.log
 
-# Sphinx documentation
-docs/_build/
+# Epydoc documentation
+docs/
 
 # PyBuilder
 target/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,10 @@
 language: python
+addons:
+  apt:
+    sources:
+      - deadsnakes
+    packages:
+      - python3.5
 services:
   - redis-server
 env:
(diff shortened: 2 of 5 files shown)


rochacbruno:
:+1:

rochacbruno:
six is already on project, so it can be used to replace invalid code ex: #8

ellisonleao:
Looks like travis doesn't support 3.5 for now https://github.com/travis-ci/travis-ci/issues/4794

rochacbruno:
Python3.5 is available from the deadsnakes repo. Add the following to your .travis.yml:

```
addons:
  apt:
    sources:
      - deadsnakes
    packages:
      - python3.5
```

ellisonleao:
what about using https://github.com/ryanhiebert/tox-travis ?

ellisonleao:
hey @rochacbruno any thoughts on this?

rochacbruno:
@ellisonleao your call! you decide! I dont have too much experience with tox.

ellisonleao:
the problem for me here is production vs dev . Tox makes it easier to test on multiple python versions locally. Travis can test on multiple versions, but we need to keep committing and checking the builds. there are some projects to simulate the builds locally (https://github.com/travis-ci/travis-build) but i found it very overengineering for what we need.

Do we need to support python 3.5 now?

rochacbruno:
we can move to WerckerCI

http://devcenter.wercker.com/docs/languages/python.html

which uses images to test https://hub.docker.com/_/python/

(maybe travis can use docker images?)

ellisonleao:
Looks like they can, not sure about the complexity

https://docs.travis-ci.com/user/docker/

PR #15, opened 2017-03-21
Make the project work both on Python2 and Python3

douglas:
Hello Bruno =)

With this commits I made the project work with Python2 and Python3 at the same time, so you don't need to maintain a different branch for it.

Besides, that, I made the following:

- Added support for Tox to help us test the project
- Added support for pipenv (https://github.com/kennethreitz/pipenv)
- Added support for EditorConfig (http://editorconfig.org)
- Some PEP8 and pylint fixes

HTH


rochacbruno:
aws

PR #27, opened 2017-07-02
Add expanded object for yaml

Sytten:
This is my implementation proposal for https://github.com/rochacbruno/dynaconf/issues/25
It's quite simple, but effective.
We could very well use that all over the project, but this would be your call @rochacbruno


dynaconf/utils/expanded_object.py
```diff
@@ -0,0 +1,8 @@
+class ExpandedObject:
```
rochacbruno: we must keep Python 2.7 compatibility so `ExpandObject(object)` needed here

dynaconf/utils/expanded_object.py
```diff
@@ -0,0 +1,8 @@
...
+    def __init__(self, **data):
+        for key, value in data.items():
+            key = key.strip().upper()
+            if isinstance(value, dict):
+                self.__dict__[key] = ExpandedObject(**value)
```
rochacbruno: maybe `setattr(self, key, ExpandedObject(**value))` is best here.

Sytten:
I will do the changes, should we use that principle everywhere or just in YAML?

dynaconf/utils/expanded_object.py
rochacbruno: I was also thinking that `ExpandedObject` may inherit from `dict` so it keep the dict methods as `get` and can be evaluated as an instance of a `dict`

rochacbruno:
if `ExpandedObject` inherits from `dict` (maybe needs to use `__getaattr__` or maybe [box](https://github.com/cdgriffith/Box) so we can make it general and not only for yaml.

Sytten:
I like this Box plugin! I refactored to use it everywhere. Since they support both the usual dict and the dot notation it should not be breaking.
the only thing I don't fully like is if the keys are not upper case, it will do something like ```settings.SERVER.host```.

rochacbruno:
@Sytten merged + I fixed the access with uppercase `settings.SERVER.HOST` will work and also `settings.SERVER.host` will work too

Sytten:
Nice! Good job :)

PR #28, opened 2017-10-04
Converted README.md to README.rst using pandoc

### LINKED ISSUE #17: Fix PyPI description
I tried to convert README.md to README.rst using `m2r`

the code is on `setup.py`

does not works, pypi page is bas formatted: https://pypi.python.org/pypi/dynaconf

I need help to convert README to proper RsT or even rewrite README in valid .rst and drop .md.

savuir:
Should fix #17 I used:
`pandoc --from=markdown --to=rst --output=README.rst README.md`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,0 +1,556 @@
+    **dynaconf** - The **dyna**\ mic **conf**\ igurator for your Python
+    Project
+
+|MIT License| |PyPI| |PyPI| |Travis CI| |Coverage Status| |Codacy grade|
+
+**dynaconf** is an OSM (**O**\ bject **S**\ ettings **M**\ apper) it can
+read settings variables from a set of different data stores such as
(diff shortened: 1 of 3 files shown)


rochacbruno:
The idea is to have a permanent solution as in

https://github.com/rochacbruno/dynaconf/blob/master/setup.py#L19

So every time we run `setup.py sdist build` the README is converted.

I would like to have the pandoc lines on setup.py, or maybe on `Makefile`

mstuttgart:
In my projects i use `.rst` and load it like this:https://github.com/mstuttgart/pycep-correios/blob/develop/setup.py#L17
It works like a charm :+1:

rochacbruno:
Fixed in https://github.com/rochacbruno/dynaconf/pull/38

PR #32, opened 2017-10-23
Added cascade look up for settings files

renzon:
#close 31

It will look on on this order for settings.py, settings.yml or settings.yaml on root directory. E.g. if you already have a settings.py it will ignore the the other 2 extensions.

### DIFF
--- a/dynaconf/default_settings.py
+++ b/dynaconf/default_settings.py
@@ -1,9 +1,15 @@
 # default proj root
 # pragma: no cover
+from os import path
+
 PROJECT_ROOT = "."
 
-# Default settings file
-SETTINGS_MODULE_FOR_DYNACONF = 'settings.py'
+for _config_file in ('settings.py', 'settings.yml', 'settings.yaml'):
+    if path.exists(path.join(PROJECT_ROOT, _config_file)):
+        SETTINGS_MODULE_FOR_DYNACONF = _config_file
+        break
+else:
+    SETTINGS_MODULE_FOR_DYNACONF = 'settings.py'
 
 # Namespace for envvars
 DYNACONF_NAMESPACE = 'DYNACONF'
--- a/README.md
+++ b/README.md
@@ -61,8 +61,13 @@ pip install dynaconf
 ```
 
 > NOTE: this project officially supports and encourages only Python 3+. Currently this is working well and tests are passing on any Python version above 2.7 but at any moment we can drop python2.x support if needed.
+# settings module
+
+If you have a settings.py, settings.yml or a settings.yaml in project's root, dynaconf will load it for you.
+Otherwise see next section.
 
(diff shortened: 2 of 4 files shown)


vladcalin:
Hello,
I think it should be specified what happens when more valid settings file is found in the root directory. (will the keys be merged? will the last valid found file be used or the first? etc)

renzon:
@vladcalin I added the description what is doing. I kept setting.py as the first one to look to not break stuff on projects already using dynaconf. But with this change, one can just create a yml on root and she will not need to defined SETTINGS_MODULE or call configure anymore, it will simply work by convention.

renzon:
I've also changed the minor version of the lib once one using the old version can upgrade to this new, but one using the new can not downgrade since the code can be relying on yaml look up by default.

rochacbruno:
@vladcalin good question!

This is related to the `default settings` and this is the `main`, the `point of entry` to read all other variables.

I think it will be hard to make a decision on which priority give to a `merging cascade`

So my opinion is, try to look for `settings.py` then `settings.{yaml|yml}` then  `settings.toml` then `settings.ini` etc...

But should exist only one default `settings` file in the root. (maybe we should fail if more than one `settings.{ext}` exists.

Then , having a default settings loaded it is easy to define a `LOADERS` or `YAML` variable inside it to have more files read.

dynaconf/default_settings.py
```diff
@@ -1,9 +1,15 @@
 # default proj root
 # pragma: no cover
+from os import path
+
 PROJECT_ROOT = "."
 
-# Default settings file
-SETTINGS_MODULE_FOR_DYNACONF = 'settings.py'
+for _config_file in ('settings.py', 'settings.yml', 'settings.yaml'):
+    if path.exists(path.join(PROJECT_ROOT, _config_file)):
```
rochacbruno: To avoid confusions, what do you think of checking if there is only one `default settings`?

There should be only one of `_config_file` if more than one found, then a `RunTimeError` is raised?

renzon: I liked your idea because it would be a bad smell having both config files. But taking DJango projects as an example, the project already has a settings.py and one could define a yaml to be used only by dynaconf. On this case if a throw an Exeception user would have to change the file name with no choice. I middle ground would be logging a warning about the presence of multiple config files. What you think?

renzon: Thinking more about it, on Django's example one would still be required to set am env var with to setting module and on this case she could also change yml file name to avoid name confusion.

renzon: A create PR #33 as an option rasing a exception so we can decide wich is better.

rochacbruno:
Implemented in https://github.com/rochacbruno/dynaconf/pull/38

Issue #19, opened 2017-03-22
Add Django extension


rochacbruno:
A Django extension when enabled should replace `django.conf.settings` with Dynaconf instance

rochacbruno:
There is a `dynaconf.contrib` with a Flask example

rochacbruno:
Right now it is possible to use Dynaconf with django in a simple way by importing it on `settings.py`

In `settings.py` of django project one can do:

```py
from dynaconf import settings

VAR = settings.VAR
DEBUG = settings.as_bool('DEBUG')
ANOTHER_VALUE = settings.get('VALUE')
...
```

But it would be perfect if explicitly reading value by value is not needed and the `django.conf.settings` object can be `patched` and replaced by `dynaconf.settings`

This way it is possible to use fresh vars, casting, and all other dynaconf features.

The expected feature is by simply doing in a `app/views.py` in django

```py
from django.conf import settings   # this imports a `patched` `settings` object which is a `Dynaconf`

def index(request):
    settings.FOO   # this value is taken from dynaconf
    settings.get_fresh('BAR')   # this value is freshly taken from dynaconf (Redis etc...)
    ...
```

Right now my POC is to do the `patch` in `manage.py` or in `wsgi.py`


![dyna_django](https://user-images.githubusercontent.com/458654/31954848-6a1a14c8-b8c5-11e7-97f5-577aac1fd6a8.png)


Django has in `django/__init__.py` a `setup()` method which is the start point for manage and also wsgi, but this method is called only when it is started, I have not found a way to override it other than `hacking python module import hooks`

rochacbruno:
Note that the idea is to **NOT TOUCH** Django's code, only **PROJECT** code, that is why `manage.py` and `wsgi.py` is the first approach.

cuducos:
And old and ugly habit of Django developers is (was?) to add something like that at the bottom of the file to bring new and external stuff to it:


```python
try:
    from whatever_settings_module import *
except ImportError:
    pass
```

So we could offer something like that in `dynaconf.contrib` – a module to be imported that behaves as an injection in the namespace… have you tried something in that sense?

Ugly as hell… but just brainstorming here ; )

renzon:
[...]

PR #33, opened 2017-10-23
Added cascade look up for settings files

renzon:
#close 31

It will look, on this order, for settings.py, settings.yml or settings.yaml on root directory.

I created this PR as an option during this discussion: https://github.com/rochacbruno/dynaconf/pull/32#pullrequestreview-71213291


rochacbruno:
@renzon take a look at the `django` extension..

The concern is, it must be a hack in `manage.py` and in `wsgi.py` and would be better to find another way to initialize it

dyna_django

rochacbruno:
depending on #19 so we fix it without conflicts

rochacbruno:
Implemented in https://github.com/rochacbruno/dynaconf/pull/38

PR #34, opened 2017-10-24
Added Django extension

rochacbruno:
IN django project use `dynaconf.contrib.django_dynaconf` as the first app

```py
# mandatory
INSTALLED_APPS = [
    'dynaconf.contrib.django_dynaconf',
    'django.contrib.admin',
    ...
]

# optional
DYNACONF_SETTINGS_MODULE = 'my_settings.yml'   # default to settings.yml in the the same folder as manage.py

DYNACONF_NAMESPACE = 'DJANGO'  # default to DYNACONF

*ANY_OTHER_DYNACONF_VARIABLE_HERE
```

Now just use `from django.conf import settings` and it will include dynaconf settings.

dyna_django2

## Caveats

Django config variables, such as MIDDLEWARES, INSTALLED_APPS etc still needs to be in `settings.py` and the dynaconf settings will be valid only for application specific settings.

## Solution 2

If it is needed to have Django settings in DYNACONF one can go to `manage.py` and `wsgi.py` and call from there `from dynaconf.contrib import django_dynaconf`

### DIFF
--- a/dynaconf/contrib/django_dynaconf/__init__.py
+++ b/dynaconf/contrib/django_dynaconf/__init__.py
@@ -1,0 +1,16 @@
+# coding: utf-8
+import django
+from dynaconf import LazySettings
+
+
+django_settings = {}
+for key in dir(django.conf.settings):
+    if key.isupper():
+        django_settings[key] = getattr(django.conf.settings, key, None)
+
+
+django_settings['SETTINGS_MODULE'] = django_settings.get(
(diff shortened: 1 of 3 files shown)


renzon:
Well, after creating a sketon on #35  to test your code I got what you mean about having vars on settings.py. Kind of ugly once one could change values there and not understand why they are not affecting the application. Would be nice to rid of them. Or maybe just fill all of them if values as: Coming from dynaconf lol

PR #35, opened 2017-10-25
Added Django Skeleton

renzon:
It is working, but only when settings is left with its vars.

Ex: I've tried to remove SECRET_KEY from settings.py and got:

  raise ImproperlyConfigured("The SECRET_KEY setting must not be empty.")
django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty.

Now I got what you mean we need to keep standard DJANGO var on settings.py, only to please the health check of the project. On your second solution, patching manage.py and wsgi.py are the var also necessary on settings.py? If not, I would go with the second solution ;)


dynaconf/contrib/django_dynaconf/apps.py
```diff
@@ -1,5 +1,5 @@
 from django.apps import AppConfig
 
 
-class BarConfig(AppConfig):
+class DJangoDynaconfig(AppConfig):
```
mazulo: I think we have a little typo in the class name

renzon: I tried I little merge with Dynaconf with Config, but maybe I've failed ;)

vladcalin: i think @mazulo is refering to "DJango..." instead of "Django"

PR #75, opened 2018-07-30
Fix in 'dynaconf/base.py' for __getitem__ method

yoyonel:
```python
def __getitem__(self, item):
        """Allow getting variables as dict keys `settings['KEY']`"""
        value = self.get(item)
        if not value:    # Problem here !
            raise KeyError('{0} does not exists'.format(item))
        return value
```
This code don't work for integer or boolean arguments/parameters (in validation step) if there are set to 0 or False.
The condition `if not value` return True if value={0, False} then raise an exception (not wanted).
A simple fix on the condition prevent this behaviour => `if value is None`


rochacbruno:
Good catch!

PR #88, opened 2018-09-27
[RFR] Merge data structures

rsnyman:
- Add the ability to merge nested structures instead of completely overwriting them

Please let me know if this looks like an acceptable solution, and what I should change to make it better.

### DIFF
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -14,7 +14,7 @@ from dynaconf.loaders import (
 )
 from dynaconf.utils.functional import LazyObject, empty
 from dynaconf.utils.parse_conf import converters, parse_conf_data, true_values
-from dynaconf.utils import BANNER, compat_kwargs, raw_logger
+from dynaconf.utils import BANNER, compat_kwargs, raw_logger, object_merge
 from dynaconf.validator import ValidatorList
 from dynaconf.utils.boxing import DynaBox
 
@@ -494,10 +494,15 @@ class Settings(object):
         :param tomlfy: Bool define if value is parsed by toml (defaults False)
         """
         value = parse_conf_data(value, tomlfy=tomlfy)
+        key = key.strip().upper()
+
+        if getattr(self, 'MERGE_ENABLED_FOR_DYNACONF', False):
+            attr = getattr(self, key, None)
+            object_merge(attr, value)
+
         if isinstance(value, dict):
             value = DynaBox(value, box_it_up=True)
 
-        key = key.strip().upper()
         setattr(self, key, value)
         self.store[key] = value
         self._deleted.discard(key)
--- a/dynaconf/default_settings.py
+++ b/dynaconf/default_settings.py
@@ -85,6 +85,9 @@ GLOBAL_ENV_FOR_DYNACONF = get('GLOBAL_ENV_FOR_DYNACONF', 'DYNACONF')
 # The default encoding to open settings files
 ENCODING_FOR_DYNACONF = get('ENCODING_FOR_DYNACONF', 'utf-8')
 
+# Merge objects on load
+MERGE_ENABLED_FOR_DYNACONF = get('MERGE_ENABLED_FOR_DYNACONF', False)
+
 # The env var specifying settings module
 ENVVAR_FOR_DYNACONF = get('ENVVAR_FOR_DYNACONF',
                           'SETTINGS_MODULE_FOR_DYNACONF')
--- a/dynaconf/utils/__init__.py
+++ b/dynaconf/utils/__init__.py
@@ -12,17 +12,19 @@ BANNER = """
 """
 
 
-def dictmerge(old, new):
+def object_merge(old, new):
     """
-    Recursively update a dict.
-    Subdict's won't be overwritten but also updated.
+    Recursively merge two data structures
     """
-    for key, value in old.items():
-        if key not in new:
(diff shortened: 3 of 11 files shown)


dynaconf/utils/__init__.py
```diff
@@ -25,6 +25,21 @@ def dictmerge(old, new):
     return new
 
 
+def deep_merge(old, new):
```
rochacbruno: There is already this function https://github.com/rochacbruno/dynaconf/blob/master/dynaconf/utils/__init__.py#L15 `dictmerge` which is used by dynaconf cli `write` commands. I think the purposes are the same. Maybe we can keep only one function for this and extend its functionality. (you implementation looks more complete)

dynaconf/default_settings.py
```diff
@@ -85,6 +85,9 @@ def start_dotenv(obj=None):
 # The default encoding to open settings files
 ENCODING_FOR_DYNACONF = get('ENCODING_FOR_DYNACONF', 'utf-8')
 
+# Merge dictionaries on load
+MERGE_CONF_FOR_DYNACONF = get('MERGE_CONF_FOR_DYNACONF', False)
```
rochacbruno: I like having this feature as optional, and this config enables it globally, what about cases where we need only specific data to be merged and others to be overwritten.

Is it a good idea to have a `super()` keyword to give a fine control over where the merge is being applied.

Example.
settings.yml
```yaml
default:
  data:
    one: 1
    two: 2
```
Now if I want to add a `three` I can

othersettings.yaml
```yaml
dynaconf_merge: true
default:
  data:
    three: 3
```

or maybe

yeatanothersettings.yaml
```yaml
default:
  data:
    dynaconf_merge: true
    four: 3
```

and at the end I have

```py
>> settings.DATA
{
   'one': 1,
   'two': 2,
   'three': 3,
   'four': 4,
}
```

The same can be applied to environment variables

```bash
$ export DYNACONF_DATA="{five = 5, six = 6, dynaconf_merge = true}"
```

If I dont supply the special key `dynaconf_merge` (we can name it just `_merge`), then the cascade will overwrite the values.

What do you think is best?

- Fine control
- global set

dynaconf/utils/__init__.py
rsnyman: Yes, I was thinking of doing that. I tried dictmerge initially, but it only handles dicts. I'll replace dictmerge with my function.

dynaconf/default_settings.py
rsnyman: I like the idea of fine control, but I'm not sure that another key in the settings file is necessarily the correct way to do it. It looks like a setting here, but when you actually look at your settings, there's no key.
Perhaps a better approach would be to start with a global setting, and then think about a good way to implement the fine control.

rochacbruno: @rsnyman agreed! the fine control doesn't block the global implementation.

If we are going to add special keywords then we need to think about the naming and test if all the file formats supports names like `_dynaconf_merge`, `_dynaconf_include` etc.

dynaconf/utils/__init__.py
rochacbruno: @RonnyPfannschmidt good point! `in_place` argument can control it. @rsnyman

rsnyman: @RonnyPfannschmidt Good point. An in-place merge is probably better.

rsnyman:
@rochacbruno Is there anything else you want me to do with this?

dynaconf/base.py
```diff
@@ -494,10 +494,17 @@ def set(self, key, value, loader_identifier=None, tomlfy=False):
         :param tomlfy: Bool define if value is parsed by toml (defaults False)
         """
         value = parse_conf_data(value, tomlfy=tomlfy)
+        key = key.strip().upper()
+
+        if 'FOR_DYNACONF' not in key and getattr(self,
```
rochacbruno: I am not sure I understand this condition, can you explain the case where `FOR_DYNACONF` key is set and should not be merged.

Also about the name of the variable `MERGE_CONF_FOR_DYNACONF` should we adopt something different?

```
ENABLE_MERGE_FOR_DYNACONF
MERGE_ENABLED_FOR_DYNACONF
```

?

rsnyman: I was experiencing some issues with keys, I'll test it again and see if it makes a difference.

As for the name, I don't mind either of those, I would just recommend they follow the same convention as the rest of the names.

rsnyman: If I remove the check for DynaConf variables, I get the following traceback:

```
...
  File "dynaconf/base.py", line 500, in set
    attr = self.get(key, default=None)
  File "dynaconf/base.py", line 231, in get
    (fresh or self._fresh or key in self.FRESH_VARS_FOR_DYNACONF) and
AttributeError: 'Settings' object has no attribute 'FRESH_VARS_FOR_DYNACONF'
```

rsnyman: Having said that, if I change line 500 in `base.py` to use `getattr` instead, it works.

```python
attr = getattr(self, key, None)
```

rochacbruno:
@rsnyman I' ll perform some manual tests on this PR this week .

rochacbruno:
Merged and added examples https://github.com/rochacbruno/dynaconf/tree/master/example/merge_configs

PR #93, opened 2018-10-16
Allow dot-traversal access to nested dictionaries.

jperras:
A simple memoized recursion has been added to `get()` if the key contains at least one dot.

The caller can choose to opt-out of this behaviour by specifying the `dotted_lookup` argument:

```python
settings('AUTH.USERNAME', dotted_lookup=False)
```

While the performance impact of this has not been quantified, the net impact in any real-world application should be minimal due to typical nesting levels, and the fact that we overwrite the memoized portion of the dotted-key lookup on each iteration.

Perhaps at some point a suite of benchmarks can be added, to which the impact of dotted-key lookups can be analyzed further.

- Avoids regressions [✓]
- Can be opted-out on a per-call basis [✓]
- Minimal performance impact [✓]
- Documented [✓]
- Tested [✓]
- Examples added [✓]

I tried to find all of the bits & pieces of docs and examples to make sure that this feature was correctly described and in the style of the project, but I wouldn't be surprised if I missed something. Please let me know!

Also: 

Closes rochacbruno/dynaconf#84

### DIFF
--- a/docs/guides/accessing_values.md
+++ b/docs/guides/accessing_values.md
@@ -41,6 +41,20 @@ settings.get('TIMEOUT', 300)
 
 > Returns the default (300) if not defined
 
+Using dotted-path lookup
+
+```python
+settings.get('AUTH.USER', 'anonymous')
+```
+
+> Returns the default ('anonymous') if not defined
+
+Explicitly disabling dotted-path lookup
+
+```python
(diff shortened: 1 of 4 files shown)


rochacbruno:
Awesome! it is in my review list :)

rochacbruno:
Merged thanks!

jperras:
Glad to help!

PR #98, opened 2018-10-22
Implements dotted-path validator name declarations.

jperras:
One is now able to write:

```python
settings.validators.register(
    Validator('REDIS',  must_exist=True,  is_type_of=dict),
    Validator('REDIS.HOST', must_exist=True, is_type_of=str),
    Validator('REDIS.PORT', must_exist=True, is_type_of=int),
)
```

Which will validate the dotted attributes as nested structures. For example, in yaml:

```yaml
DEFAULT:
    REDIS:
        HOST: localhost
        PORT: 1234
```

This necessitated a slight but non-negligible change in the implementation of `Settings.exists()`, which previously did a shallow check of loaded data. It has now been updated to perform a `Settings.get()` of the key in question to unify the behaviour of `exists()` and `get()`, and compares that to a newly defined sentinel value to ensure `None` values do not cause a false negative result.

- New tests and assertions have been added to cover the new functionality.
- Docs have been updated to show an example of the nested validator name definition in action.

Closes rochacbruno/dynaconf#85.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ test_examples:
 	@cd example/flask_with_json;pwd;flask routes | grep -c flask_with_json
 	@cd example/flask_with_ini;pwd;flask routes | grep -c flask_with_ini
 	@cd example/validators/with_python/;pwd;python app.py | grep -c validator
+	@cd example/validators/with_toml/;pwd;dynaconf validate
 	@cd example/toml_with_secrets/;pwd;python program.py | grep -c My5up3r53c4et
 	@cd example/envs;pwd;python app.py
 	@cd example/django_example/;pwd;python manage.py test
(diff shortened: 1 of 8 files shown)


jperras:
I updated the PR (1f5666a) with some additional unit tests and an explicit `__eq__` dunder for the `Missing` singleton, to hopefully avoid some ambiguous issues in the future if it is ever misused.

rochacbruno:
Merged manually in https://github.com/rochacbruno/dynaconf/commit/cc03aa0520b9e811df866c28c96566640fc217ba

jperras:
@rochacbruno Was the manual merge because I did something wrong, or just because it was easier to do it that way due to the additional tests you added in?

rochacbruno:
@jperras just because I wanted to add some `example/validators` and test locally before merging.

rochacbruno:
@jperras I wanted to test specially this: https://github.com/rochacbruno/dynaconf/blob/master/example/validators/with_toml/dynaconf_validators.toml#L7 and it works thanks!!!

jperras:
Ah ok! Glad to help!

PR #103, opened 2018-10-30
Handle nested includes in settings files.

### LINKED ISSUE #86: RFC: Allow multiple settings files to be merged into one.
Targeting plugin based applications take as example:

Configuration for general program is in `export SETTINGS_FOR_DYNACONF=/etc/program/settings.ext` (`ext` being any of the supported formats yaml, json, ini, toml etc)

That same program has plugins and plugin stores its settings under different path like `/etc/plugin_name/settings.yaml`

## RFC: Allow `include` of settings files in different formats.

Then `/etc/program/settings.yaml`

```yaml
DEFAULT:
    DYNACONF_INCLUDE:
        - /etc/plugin_name/settings.yaml
        - /etc/other_plugin_name/settings.toml
        - /etc/settings_folder/* 
``` 

or toml

```toml
[default]
dynaconf_include = ["path1", "path2"]
```

or json

```json
{
   "default": {
        "dynaconf_include": ["path1", "path2"]
    }
}
```

then All the loaders will take the special `DYNACONF_INCLUDE` key then load values from each of the files listed using the normal pipeline.

### Requirements
- [x] Backwards compatibility
- [x] Multiple [envs] should be considered
- [x] Preserve the ordering
- [x] Nested includes should have a maximum recursion level
- [x] When the path is a directory, glob should be used and load in alpha order
- [x] All file formats should be supported, the functionality should be in BaseLoader

jperras:
## Summary

A settings file can include a `dynaconf_include` stanza, whose exact syntax will depend on the type of settings file (json, yaml, toml, etc) being used.

When loaded, the files located at the (relative or absolute) paths in the `dynaconf_include` key will be parsed, in order, and override any base settings that may exist in your current configuration.

The paths can be relative to the base `settings.(toml|yaml|json|ini|py)` file, or can be absolute paths.

The idea here is that plugins or extensions for whatever framework or architecture you are using can provide their own configuration values when necessary.

## Example Usage

```toml
[default]
dynaconf_include = ["/absolute/path/to/plugin1.toml", "relative/path/to/plugin2.toml"]
DEBUG = false
SERVER = "www.example.com"
```

It is also possible to specify glob-based patterns:

```toml
[default]
dynaconf_include = ["configurations/*.toml"]
DEBUG = false
SERVER = "www.example.com"
```

## Details/Notes

Currently, only a single level of includes is permitted to keep things simple and straightforward. Some of the work for multiple levels of nested includes has already been prepared, e.g. ensuring that we never load the same settings file twice to prevent an infinite recursion.

@rochacbruno There is currently an issue when attempting to load nested configuration files of different types, e.g. `dynaconf_include = ["plugin1.toml", "plugin1.json"]`. It appears as though the base `settings.toml` data gets *re-read* after it has been initially processed, and then overrides data that has been loaded up with `plugin1.toml`, making it seem like this latter was never processed. I've attempted to track down the cause of this but have come away empty handed, and would appreciate some advice!

The above incorrect behaviour is exposed in the `tests/test_nested_loading.py::test_load_nested_different_types` test function.

## References 

Closes #86.


rochacbruno:
@jperras great! there is a little PEP8 error to be fixed, then I can merge it! 

* investigating the problem you mentioned.

jperras:
@rochacbruno Thanks! I attempted to trace it down frame by frame, but there's a fair bit of indirection with the various loadings that occur, e.g. `.env` functionality. I'm also happy to write some more tests once the issue I indicated has been resolved.

rochacbruno:
Manually merged to include some fixes,
thanks @jperras

rochacbruno:
commit https://github.com/rochacbruno/dynaconf/commit/59b86498146047ca89ba6833af92deb985f32431

jperras:
Awesome, thanks!

PR #113, opened 2018-11-29
fixed infinite recursion caused by copy()

mspinelli:
`DynaBox` was previously inheriting `Box`'s `__copy__()` and `copy()` methods. These methods are here from class `Box`:
```
def copy(self):
    return self.__class__(super(self.__class__, self).copy())

def __copy__(self):
    return self.__class__(super(self.__class__, self).copy())
```
Due to the way super works and is coded in `Box` this causes infinite recursion whenever an instance of `DynaBox` calls `copy`. This happens for instance when Django reads a settings stored as a table in `settings.toml`. It imports it as a DynaBox but expects it to have the duck typing behavior of a dictionary (and thus the copy operation). So the fix is to override these two methods and specify specifically to call `Box`'s implementation. Alternatively `super(dict, self).copy()` works too but I thought it makes more sense to reference the implementation that DyanBox is inheriting from directly.

### DIFF
--- a/dynaconf/utils/boxing.py
+++ b/dynaconf/utils/boxing.py
@@ -21,6 +21,12 @@ class DynaBox(Box):
             n_item = item.lower() if item.isupper() else item.upper()
             return super(DynaBox, self).__getitem__(n_item, *args, **kwargs)
 
+    def __copy__():
+        return self.__class__(super(Box, self).copy())
+
+    def copy():
+        return self.__class__(super(Box, self).copy())
+
     def get(self, item, default=None, *args, **kwargs):
         value = super(DynaBox, self).get(item, default, *args, **kwargs)
         if value is None or value == default:


dynaconf/utils/boxing.py
```diff
@@ -21,6 +21,12 @@ def __getitem__(self, item, *args, **kwargs):
             n_item = item.lower() if item.isupper() else item.upper()
             return super(DynaBox, self).__getitem__(n_item, *args, **kwargs)
 
+    def __copy__():
```
rochacbruno:
```suggestion
    def __copy__(self):
```

dynaconf/utils/boxing.py
```diff
@@ -21,6 +21,12 @@ def __getitem__(self, item, *args, **kwargs):
...
+        return self.__class__(super(Box, self).copy())
+
+    def copy():
```
rochacbruno:
```suggestion
    def copy(self):
```

rochacbruno:
```

 ✔   ~/Projects/python/dynaconf/example/django_example   master 
 $ ipython                                                            

In [2]: import dynaconf.contrib.django_dynaconf                                                                                                

In [3]: from django.conf import settings                                                                                                       

In [4]: type(settings)                                                                                                                         
Out[4]: dynaconf.base.LazySettings

In [5]: settings.DATABASES                                                                                                                     
Out[5]: <Box: {'default': {'ENGINE': 'django.db.backends.sqlite3', 'NAME': '/home/rochacbruno/Projects/python/dynaconf/example/django_example/db.sqlite3'}}>

In [6]: settings.DATABASES.copy()                                                                                                              
---------------------------------------------------------------------------
RecursionError                            Traceback (most recent call last)
<ipython-input-6-3419f44ef903> in <module>
----> 1 settings.DATABASES.copy()

~/Projects/python/dynaconf/venv/bin/box.py in copy(self)
    397 
    398     def copy(self):
--> 399         return self.__class__(super(self.__class__, self).copy())
    400 
    401     def __copy__(self):

... last 1 frames repeated, from the frame below ...

~/Projects/python/dynaconf/venv/bin/box.py in copy(self)
    397 
    398     def copy(self):
--> 399         return self.__class__(super(self.__class__, self).copy())
    400 
    401     def __copy__(self):

RecursionError: maximum recursion depth exceeded while calling a Python object
```

PR #123, opened 2019-03-01
Add support for environments in Vault

rsnyman:
We want to be able to use a single Vault server for multiple environments. Currently, Dynaconf does not support multiple environments in Vault.
This PR adds the working environment into the path, so that the same Vault server can be used for multiple environments.

- Converted `os.path.join()` to `'/'.join()` as Vault only uses `/` and this would cause Dynaconf on Windows to provide an incorrect path to Vault.

### DIFF
--- a/dynaconf/loaders/vault_loader.py
+++ b/dynaconf/loaders/vault_loader.py
@@ -1,12 +1,11 @@
 # docker run -e 'VAULT_DEV_ROOT_TOKEN_ID=myroot' -p 8200:8200 vault
 # pip install hvac
 
-import os
 from dynaconf.utils.parse_conf import parse_conf_data
 
 try:
     from hvac import Client
-except ImportError as e:
+except ImportError:
     raise ImportError(
         "vault package is not installed in your environment. "
         "`pip install dynaconf[vault]` or disable the vault loader with "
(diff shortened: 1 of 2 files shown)


dynaconf/loaders/vault_loader.py
```diff
@@ -17,6 +16,28 @@
...
+    :param env: settings env default='DYNACONF'
+    :return: a list of working environments
+    """
+    # add the [default] env
+    env_list = [obj.get('DEFAULT_ENV_FOR_DYNACONF')]
+    # compatibility with older versions that still uses [dynaconf] as
+    # [default] env
+    global_env = obj.get('GLOBAL_ENV_FOR_DYNACONF')
+    if global_env not in env_list:
+        env_list.append(global_env)
```
rochacbruno: I think global should be the last item on the list, so it can be read later and override all previous environments.

rochacbruno:
Also there is this error on Travis CI https://travis-ci.org/rochacbruno/dynaconf/jobs/500591773

rochacbruno:
I am curious about this use case, are you guys changing environment on the fly? using `settings.setenv/using_env` ?

In other scenarios I have only used the variable exporting different path example:

```
export VAULT_PATH_FOR_DYNACONF=/secret/data/SOMETHINGELSE
```

dynaconf/loaders/vault_loader.py
rsnyman: I basically just copied this from `dynaconf/loaders/base.py` (line 63):

```python
        # add the [default] env
        env_list = [self.obj.get('DEFAULT_ENV_FOR_DYNACONF')]

        # compatibility with older versions that still uses [dynaconf] as
        # [default] env
        global_env = self.obj.get('GLOBAL_ENV_FOR_DYNACONF')
        if global_env not in env_list:
            env_list.append(global_env)

        # add the current [env]
        if self.env not in env_list:
            env_list.append(self.env)

        # add the [global] env
        env_list.append('GLOBAL')
```

rsnyman:
Yes, we're changing the environment on the fly using `ENV_FOR_DYNACONF`. Having do use both `ENV_FOR_DYNACONF` and `VAULT_PATH_FOR_DYNACONF` is a bit burdensome (especially if people don't get the Vault path right).

I'm busy looking into the error in Travis. I can reproduce it, but I am not sure why it is happening.

Issue #120, opened 2019-02-21
Use `__slots__` to optimize settings object


rochacbruno:
SSIA

`__slots__`

Issue #22, opened 2017-03-22
Caching on execute_loaders


rochacbruno:
Currently, all loaders are executed again every time we 
- switch namespaces using `settings.using_namespace` or `settings.namespace` 
- when a variable is marked as `always_fresh` 
- when using `settings.get('var', fresh=True) 

For the `fresh=True` and `always_fresh` it makes sense to execute loaders again as it is demanded to get the fresh variable (lets say from redis or another database).

The problem is that it may cause overhead (I guess not too much) but in any case we can improve by.

- writing more namespace tests (with more switching and `fresh` examples)
- implement a cache on `execute_loaders` so files and databases are not accessed twice when we are only switching namespaces and not wanting to load all again.

Part of this is already implemented in `loaded_by_loaders` dictionary, we just need to use this as a cache point.

cinnabot:
Hi, I have a suggestion to help solving this issue!



The issue [Allow separation of projects vars via namespace](https://github.com/rochacbruno/dynaconf/issues/1) is possibly related to this one.




Please let me know if this was useful:
- [Yes, this was useful](https://www.getcinnamon.io/posts/114dcea6-1ee7-4f9a-8579-90cdf6eaad01/rate/?rate=like)
- [No, this wasn't useful](https://www.getcinnamon.io/posts/114dcea6-1ee7-4f9a-8579-90cdf6eaad01/rate/?rate=dislike)

Cheers,
Cinnabot.


PS: Liked my suggestions? Install me in your repository! Visit: [getcinnamon.io](https://www.getcinnamon.io).

PR #124, opened 2019-03-08
Fix exponential slow down when loader is run multiple times

psav:
* When using context managers, the loader is invoked each time.
  This was slowing down in an exponential manner each time the it was run.
  The eventual cause of this was down to an attribute being used as a list.
  The object merge dutifully tried to expand this item out again and again
  even in the case that the list was a single item, resulting in [item],
  becoming [item, item]. The next time the merge was run, this process was
  run again, but for each item in the list. In this particular instance
  the list was identical, it meant that the list grew exponentially.
* This fix is a short optimization that checks to see if the old list
  is identical to the new list. In which case, there is no merge to complete
  so we simply return.

### DIFF
--- a/dynaconf/utils/__init__.py
+++ b/dynaconf/utils/__init__.py
@@ -17,6 +17,8 @@ def object_merge(old, new):
     Recursively merge two data structures
     """
     if isinstance(old, list) and isinstance(new, list):
+        if old == new:
+            return
         for item in old[::-1]:
             new.insert(0, item)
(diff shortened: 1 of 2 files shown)


rochacbruno:
Thanks @psav there other places where we can improve performance and caching.

Like #120 use of `__slots__` and also implementing a configurable caching for variables. #22

psav:
Sure thing! This one was just biting me hard!

Issue #111, opened 2018-11-15
MERGE_ENABLED causes application not to start


bmbouter:
Pulp uses dynaconf for [its configuration](https://docs.pulpproject.org/en/3.0/nightly/installation/configuration.html). Pulp's django webserver for example, that will start successfully w/ the django extension for Dynaconf.

However on the terminal when I: `export MERGE_ENABLED_FOR_DYNACONF=true` I get the following traceback upon starting the djagno webserver:

```
(pulp) [vagrant@pulp3 ~]$ pulp-manager runserver
Unhandled exception in thread started by <function check_errors.<locals>.wrapper at 0x7fe95e0e5a60>
Traceback (most recent call last):
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.6/site-packages/django/utils/autoreload.py", line 225, in wrapper
    fn(*args, **kwargs)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.6/site-packages/django/core/management/commands/runserver.py", line 109, in inner_run
    autoreload.raise_last_exception()
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.6/site-packages/django/utils/autoreload.py", line 248, in raise_last_exception
    raise _exception[1]
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.6/site-packages/django/core/management/__init__.py", line 337, in execute
    autoreload.check_errors(django.setup)()
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.6/site-packages/django/utils/autoreload.py", line 225, in wrapper
    fn(*args, **kwargs)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.6/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/home/vagrant/.virtualenvs/pulp/lib64/python3.6/site-packages/django/apps/registry.py", line 93, in populate
    "duplicates: %s" % app_config.label)
django.core.exceptions.ImproperlyConfigured: Application labels aren't unique, duplicates: admin
```

adevore:
I've encountered something similar. I think I traced it to how LazySettings handles default values. Values are first set in the LazySettings constructor via `self._kwargs = kwargs`. Later, the default_loader function runs LazySettings.set(), which proceeds to merge each kwarg with itself.

rochacbruno:
In progress on #135

rochacbruno:
[...]

PR #135, opened 2019-03-26
Add merge support per variable using cast style

rochacbruno:
When you have existing variables of type dict or list and you want to contribute to existing data instead of overriding
it is possible to use the special `@merge` notation.

Existing `settings.toml`
```toml
[default]
colors = ['yellow', 'blue', 'red']
```

Now you want to add `green` color to that existing list via environment variable:

```bash
export DYNACONF_COLORS="@merge ['green']"
```

When you access the variable you will get:

```py
from dynaconf import settings
assert settings.COLORS == ['green', 'yellow', 'blue', 'red']
```

But, what if green was already on the list? Dynaconf default behavior is to duplicate it like:

```toml
[default]
colors = ['yellow', 'blue', 'red', 'green']
```
```bash
export DYNACONF_COLORS="@merge ['green']"
```
```py
assert settings.COLORS == ['green', 'yellow', 'blue', 'red', 'green']
```

See the double `green`? how to avoid that? use **@merge_unique** which will not allow duplicated items, it will remove the duplicate from the existing list and add the new one in the new defined order.

```bash
export DYNACONF_COLORS="@merge_unique ['green']"
```
```py
assert settings.COLORS == ['yellow', 'blue', 'red', 'green']
```

The **@merge** also works for dictionaries:

```toml
[default]
database = {host="server.com", user="admin"}
```
```bash
export DYNACONF_DATABASE='@merge {password="s3cr4t"}'
```
```py
assert settings.DATABASE == {'host': 'server.com', 'user': 'admin', 'password': 's3cr4t'}
```

> NOTE: for dictionaries `@merge_unique` does not have any effect as it will by default exclude existing duplicate keys and replace with the newer ones.

#### Known caveats

The merge functionality works only for the **first** level keys on a dictionary and only **first** level items on a list.

In other words:

if I do `export DYNACONF_COLORS='@merge ["blue", "red", "green"]'`

```yaml
default:
  colors:
    - yellow
```

the exported variable will work and will merge to `colors` because it is the first level list on the `default` env and `settings.COLORS` is now `["blue", "red", "green", "yellow"]`.

However if I do if I do `export DYNACONF_COLORS='@merge {basic=["blue", "red", "green"]}'`

```yaml
default:
  colors:
    basic:
      - yellow
    compound:
      - fucsia
```

Now it is not merging the nested elements and you loose existing `yellow` value having `settings.BASIC.colors` as `["blue", "red", "green"]` and you kept `settings.COLORS.compound` unchanged.


rochacbruno:
Hi @psav / @rsnyman I am planning to introduce this new thing and I would like to get your comments on it as you both worked on the `merge` feature.

I am adding this because of the need to solve an existing Django issue #111 and this way we can keep `MERGE_ENABLED_FOR_DYNACONF` disabled and merge only specific **first** level keys.

What do you guys think of this solution? any concern? any performance test you can do?

rochacbruno:
@psav @rsnyman 

Future plan is to add something more clever for merge, without using the `@merge` notation, for future the idea is something like:

having 

```toml
[default]
database = {host="server.com", user="admin"}
```
I can do:

```bash
export DYNACONF_DATABASE='{dynaconf_merge=true, password=1234}'
```

And the presence of `dynaconf_merge=true` will mark that dictionary as a candidate for merging. (the same way we currently do with `dynaconf_include` token.

The same for lists `export DYNACONF_COLORS='["dynaconf_merge", "green"]'` and when `dynaconf_merge` item is present, the list is candidate for merging.

I would appreciate your input on this other idea also.

rsnyman:
@rochacbruno I'll take a look and let you know.

rochacbruno:
@bmbouter @dkliban your input would be also very helpful as you are also pontential end-users of this feature

bmbouter:
@rochacbruno the feature as described in [this comment](https://github.com/rochacbruno/dynaconf/pull/135#issuecomment-476880207) makes a lot of sense to me.

rochacbruno:
@bmbouter @dkliban @psav @rsnyman 

Closing in favor of #139 (better implementation)

PR #162, opened 2019-04-17
Skip reloading envs for validators that only apply to current_env

janw:
Hey @rochacbruno 👋 

### Problem statement

So I've been using Dynaconf extensively for the past few days, and something I noticed with adding more Validators to my settings was a significantly increased delay in validation. With debug logging enabled it became obvious that for every Validator instance added to the settings, all loaders were re-run when doing `settings.validators.validate()`, i.e. eloading the entire environment once per Validator. 

That makes sense when a Validator instance has the `env=` param set to anything not equal to `settings.current_env` but not when the Validator is defined with `env=None` or equal to `settings.current_env`. In that case the `using_env` contextmanager would superfluously reload the current environment.

### Proposal

In this PR I separated the internal validation logic in `Validator.validate()` into `Validator._validate_items()` and added a special case that would skip the `settings.using_env` contextmanager and go directly into the validation logic.

### Benefit

In a quick-and-dirty trial, I was able to improve the duration to go through 100 Validator instances from about 3 seconds down to less than .1 seconds. Afaics there's no down-side for changing the Validator's behavior this way.

```python
from datetime import datetime
from dynaconf import settings, Validator

vals = (Validator("SOMEVAR", ne="") for _ in range(100))

start = datetime.now()
settings.validators.register(*vals)

print(datetime.now()-start)
```

Cheers!

### DIFF
--- a/dynaconf/validator.py
+++ b/dynaconf/validator.py
@@ -100,7 +100,10 @@ class Validator(object):
                 return
 
         # If only using current_env, skip using_env decoration (reload)
-        if len(self.envs) == 1 and self.envs[0] == settings.current_env:
+        if (
+            len(self.envs) == 1 and
+            self.envs[0].upper() == settings.current_env.upper()
+        ):
             self._validate_items(settings, settings.current_env)
             return
 
--- a/dynaconf/validator.py
+++ b/dynaconf/validator.py
@@ -99,53 +99,62 @@ class Validator(object):
                 # if when is invalid, return canceling validation flow
                 return
 
+        # If only using current_env, skip using_env decoration (reload)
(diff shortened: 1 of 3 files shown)


janw:
By the way: if you want me to add some unittests specifically assuring no unecessary calls to `settings.setenv` are made, I can add some using [unittest.mock](https://docs.python.org/3/library/unittest.mock.html) or its pytest integration through pytest-mock. Let me know if you want that dependency to be added here.

rochacbruno:
@janw yep  more tests would be nice! you can add `mock` requirements to https://github.com/rochacbruno/dynaconf/blob/master/requirements_dev.txt

I'll review/merge soon and add this to next 2.0.1 release.

Thanks :)

PR #176, opened 2019-04-23
Fix `False` not being an acceptable env

janw:
When using `ENVVAR_PREFIX_FOR_DYNACONF=False` in LazySettings in combination with for example a TOML settings file, the base loader will raise an unexpected exception trying to `.lower`case the boolean. This MR appends `DYNACONF` as compatibility global_env instead when the `ENVVAR_PREFIX_FOR_DYNACONF` is `False`.

**Steps to reproduce the issue:**

1. Create TOML settings file `settings.toml` with the following contents:
  ```toml
  [global]
  var = 1
  ```
2. Try instantiating with `ENVVAR_PREFIX_FOR_DYNACONF=False`, and accessing an arbitrary value from the settings like so:

```python
settings = LazySettings(
    ENVVAR_PREFIX_FOR_DYNACONF=False, 
    SETTINGS_FILE_FOR_DYNACONF="settings.toml"
)
assert settings.VAR == 1
```

I added testcases to ensure this undefined behavior is properly prevented in the future. 

Cheers! 👋

### DIFF
--- a/dynaconf/loaders/base.py
+++ b/dynaconf/loaders/base.py
@@ -68,7 +68,7 @@ class BaseLoader(object):
 
         # compatibility with older versions that still uses [dynaconf] as
         # [default] env
-        global_env = self.obj.get("ENVVAR_PREFIX_FOR_DYNACONF")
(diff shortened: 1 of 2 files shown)


rochacbruno:
good catch! :100:

Issue #211, opened 2019-08-24
[RFC] Add method to load_keys


kedark3:
**Problem**
Suppose we have a secret store that has layout like this:
![image](https://user-images.githubusercontent.com/8180469/63630714-f763a580-c5eb-11e9-957b-4fe469a9962b.png)

I would like to have a method to list all the keys. 

**Solution**
I have tried following piece of code in Vault_loader.py as:
```
def load_keys(obj, path):
    client = get_client(obj)
    return client.list('/secret/metadata/{}'.format(path))
```
It uses `/secret/metadata` endpoint for Vault and returns the dict that also contains keys. 

**Sample output:**
```
{'request_id': '<redacted>',
 'lease_id': '',
 'renewable': False,
 'lease_duration': 0,
 'data': {'keys': ['secret1', 'secret2', 'secret3']},
 'wrap_info': None,
 'warnings': None,
 'auth': None}

```
I can then do 
```
vault_loader.load_keys(settings, 'sample/')['data']['keys']                                                    
['secret1', 'secret2', 'secret3']
```

**Describe alternatives you've considered**
IMHO, It only fits to use this solution because we have number of different credentials we want to store and retrieve. Another solution would be to dump all credentials under one secret in JSON format like this 
![image](https://user-images.githubusercontent.com/8180469/63630822-e6fffa80-c5ec-11e9-8108-1668fffb14f0.png)

The problem with that is if we have a lot of credentials(keys) in one secret(or environment) then updating one key would mean that I have to load entire data and then update relevant key and write the entire data again. 

**Additional context**
You may contact me on Google Chat/email(Red Hat) if you want to discuss our use case. 
I am fairly new to Vault/Dynaconf, excuse if my understanding is different that how this all is intended to be used.

VaultVulp:
I'm a (sort of :)) active Vault and Dynaconf user and would like to know a little bit more about your use case.

Why do you need to get Vault internal metadata through Dynaconf via additional function? As far as I understand (@rochacbruno can correct me if I'm wrong) - Dynaconf helps us to get one source of truth instead of reading inconsistent data in different modules from different sources such as os.get_env(), config files, .py-file, etc. All this source will be hidden behind the same Dynaconf API: `settings.PATH.KEY`. But in your PR #214 you provide a new API call specifically for Vault.

And still, if you split your secrets in a way described here you won't be able to read this values directly via Dynaconf, due to how `vault_loader.load` works.

[...]

PR #214, opened 2019-08-27
Adding new feature to address issue #211

kedark3:
Addresses Issue #211 

This also removes restriction placed on `env` name to not include `_` characters.

### DIFF
--- a/dynaconf/loaders/vault_loader.py
+++ b/dynaconf/loaders/vault_loader.py
@@ -130,3 +130,17 @@ def write(obj, data=None, **kwargs):
     path = "/".join([obj.VAULT_PATH_FOR_DYNACONF, obj.current_env.lower()])
     client.secrets.kv.create_or_update_secret(path, secret=data)
     load(obj)
+
+
+def load_keys(obj, path):
+    """load all the keys specified at gived path in vault.
+
+    :param obj: settings object
+    :param path: path to the vault secrets
+    :return: list containing all the keys at the given path
+    """
+    client = get_client(obj)
+    try:
+        return client.list('/secret/metadata/{}'.format(path))['data']['keys']
+    except TypeError:
+        return []
--- a/dynaconf/loaders/vault_loader.py
+++ b/dynaconf/loaders/vault_loader.py
@@ -132,15 +132,22 @@ def write(obj, data=None, **kwargs):
     load(obj)
 
 
-def load_keys(obj, path):
-    """load all the keys specified at gived path in vault.
+def list_envs(obj, path=""):
+    """
+    This function is a helper to get a list of all the existing envs in
+    the source of data, the use case is:
+        existing_envs = vault_loader.list_envs(settings)
+        for env in exiting_envs:
+            with settings.using_env(env):  # switch to the env
+            # do something with a key of that env
 
     :param obj: settings object
     :param path: path to the vault secrets
     :return: list containing all the keys at the given path
     """
     client = get_client(obj)
+    path = path or obj.get("VAULT_PATH_FOR_DYNACONF")
     try:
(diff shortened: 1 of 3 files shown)


dynaconf/loaders/vault_loader.py
```diff
@@ -130,3 +130,17 @@ def write(obj, data=None, **kwargs):
     path = "/".join([obj.VAULT_PATH_FOR_DYNACONF, obj.current_env.lower()])
     client.secrets.kv.create_or_update_secret(path, secret=data)
     load(obj)
+
+
+def load_keys(obj, path):
```
rochacbruno: What do you think about naming it `list_envs` ?
```suggestion
def list_envs(obj, path):
```

In that way we can implement the same thing to all the loaders, so we make it uniform, for example, in a future PR we can add `yaml_loader.list_envs` and this will return the list of all the first level keys in the yaml file defined by `path` argument.

dynaconf/loaders/vault_loader.py
```diff
@@ -130,3 +130,17 @@ def write(obj, data=None, **kwargs):
...
+    """load all the keys specified at gived path in vault.
+
+    :param obj: settings object
+    :param path: path to the vault secrets
+    :return: list containing all the keys at the given path
+    """
```
rochacbruno: Please add some lines to docstring stating the reason for that function to exist and provide a use case, so in future we will know why that function exists as it will not be in use by dynaconf itself.

```suggestion
    ...
    This function is a helper to get a list of all the existing envs in the source of data, 
    the use case is:
        existing_envs = vault_loader.list_envs(settings)
        for env in exiting_envs:
            with settings.using_env(env):  # switch to the env
                # do something with a key of that env
        
    """
```

dynaconf/loaders/vault_loader.py
rochacbruno: Lets make `path` optional? if not passed then it will be taken from object itself?
```suggestion
def list_envs(obj, path=None):
    ...
    path = path or obj.get('VAULT_PATH_FOR_DYNACONF')
```

example/vault/vault_example.py
```diff
@@ -1,8 +1,11 @@
...
+from dynaconf.loaders.vault_loader import load_keys
 
 print(settings.SECRET)  # noqa
 # >>> 'vault_works'
 
 
 with settings.using_env("dev"):
     assert settings.SECRET == "vault_works_in_dev"
+
+assert set(load_keys(settings, "test/")) == set(["default", "dev"])
```
rochacbruno: I don't remember where the `test/` path have been defined.. by default the path would be `dynaconf/` no?

rochacbruno: yes I am right ans actually there is an assertion error in https://dev.azure.com/rochacbruno/dynaconf/_build/results?buildId=189&view=logs

For some reason the CI did not catches the problem.

```
raceback (most recent call last):
  File "vault_example.py", line 11, in <module>
    assert set(load_keys(settings, "test/")) == set(["default", "dev"])
AssertionError
1
```

dynaconf/loaders/vault_loader.py
kedark3: I like it. I will add that shortly.

dynaconf/loaders/vault_loader.py
kedark3: Sure thing.

dynaconf/loaders/vault_loader.py
kedark3: Got it. I will check how it behaves.

example/vault/vault_example.py
kedark3: It did pass locally for me, let me double check this.

Issue #218, opened 2019-09-03
[RFC] Allow first level keys to be lower-case by turning _store to a DynaBox


rochacbruno:
This comes from issue #200 and PR #217 

by @austinnichols101

> the internal _store uses a different container (dict) for the top level keys and the sub-keys (dynabox).
> 
> If _store was implemented using a dynabox then the keys could be stored using their original case and this opens up the possibility of a flag to use case-specific keys.

So we need to try moving `LazySettings._store` to a DynaBox and see what happens with having case-less first level keys.

rochacbruno:
_store is now a DynaBox, but the restriction for having first level keys upper case will be kept, it would be too much work change it now.

At any point anyone can directly insert variables to `_store` bypassing the `set` process.

Instead of `settings.set('key', 'value')` which will force key to be upper case, one can use `settings.store['key'] = 'value'` that will work, but bypasses all the logic of `set` which can deal with dotted set, fresh vars, deleted vars etc..

PR #219, opened 2019-09-03
Adds `from_env` method and change `_store` to be a `DynaBox`

rochacbruno:
Ref #218 
`settings._store` is now a `DynaBox` and that allows first level keys to be lower case.

Ref #211 
`settings` now has a `from_env` method to allow switching branches without side effects.

## Switching working environments

You can switch between existing environments using:

- `from_env`: (**recommended**) Will create a new settings instance pointing to defined env.
- `setenv`: Will set the existing instance to defined env.
- `using_env`: Context manager that will have defined env only inside its scope.

### from_env

> **New in 2.0.5**

Return a new isolated settings object pointing to specified env.

Example of settings.toml::

```ini
[development]
message = 'This is in dev'
foo = 1
[other]
message = 'this is in other env'
bar = 2
```

Program::

```py
>>> from dynaconf import settings
>>> print(settings.MESSAGE)
'This is in dev'
>>> print(settings.FOO)
1
>>> print(settings.BAR)
AttributeError: settings object has no attribute 'BAR'
```

Then you can use `from_env`:

```py
>>> print(settings.from_env('other').MESSAGE)
'This is in other env'
>>> print(settings.from_env('other').BAR)
2
>>> print(settings.from_env('other').FOO)
AttributeError: settings object has no attribute 'FOO'
```

The existing `settings` object remains the same.

```py
>>> print(settings.MESSAGE)
'This is in dev'
```

You can assign new settings objects to different `envs` like:

```py
development_settings = settings.from_env('development')
other_settings = settings.from_env('other')
```

And you can choose if the variables from different `envs` will be chained and overridden in a sequence:

```py
all_settings = settings.from_env('development', keep=True).from_env('other', keep=True)

>>> print(all_settings.MESSAGE)
'This is in other env'
>>> print(all_settings.FOO)
1
>>> print(all_settings.BAR)
2
```

The variables from [development] are loaded keeping pre-loaded values, then the variables from [other] are loaded keeping pre-loaded from [development] and overriding it.

### setenv

Will change `in_place` the `env` for the existing object.

```python
from dynaconf import settings

settings.setenv('other')
# now values comes from [other] section of config
assert settings.MESSAGE == 'This is in other env'

settings.setenv()
# now working env are back to previous
```

### using_env

Using context manager

```python
from dynaconf import settings

with settings.using_env('other'):
    # now values comes from [other] section of config
    assert settings.MESSAGE == 'This is in other env'

# existing settings back to normal after the context manager scope
assert settings.MESSAGE == 'This is in dev'
```

### DIFF
--- a/docs/guides/advanced_usage.md
+++ b/docs/guides/advanced_usage.md
@@ -24,29 +24,117 @@ Now you can use `export MYPROGRAM_FOO=bar` instead of `DYNACONF_FOO=bar`
 
 ## Switching working environments
 
-To switch the `environment` programatically you can use `setenv` or `using_env`.
+You can switch between existing environments using:
+
+- `from_env`: (**recommended**) Will create a new settings instance pointing to defined env.
+- `setenv`: Will set the existing instance to defined env.
+- `using_env`: Context manager that will have defined env only inside its scope.
+
+### from_env
+
+> **New in 2.0.5**
+
+Return a new isolated settings object pointing to specified env.
+
+Example of settings.toml::
+
+```ini
+[development]
+message = 'This is in dev'
+foo = 1
+[other]
+message = 'this is in other env'
+bar = 2
+```
+
+Program::
+
+```py
+>>> from dynaconf import settings
+>>> print(settings.MESSAGE)
+'This is in dev'
+>>> print(settings.FOO)
+1
+>>> print(settings.BAR)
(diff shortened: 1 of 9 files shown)


rochacbruno:
@austinnichols101

rochacbruno:
@kedark3 I guess this `from_env` will be useful for your use case.

Issue #193, opened 2019-05-15
[RFC] Allow alternative order of file search


PyUser5:
**Describe the bug**
Wenn dynaconf is used from a python script (invoked directly via command-line using a python shell) newer versions (>=2.0) use the path-traversal from the scripts directory and not from the current path as the old version used to do.

**To Reproduce**

1. Having the following folder structure
```
.                                                                                                                                                                                                                 
├── test1                                                                                                                                                                                                         
│   ├── pythonscript                                                                                                                                                                                              
│   └── settings.toml                                                                                                                                                                                             
└── test2                                                                                                                                                                                                         
    └── settings.toml    
```

2. Having the following config files:

<details>
<summary> Config files </summary>

**test1/settings.toml**
```toml
[default]
TESTVAL = "test1"

```

and

**test2/settings.toml**
```toml
[default]
TESTVAL = "test2"
```

</details>

3. Having the following app code:

<details>
<summary> Code </summary>

**test1/pythonscript**
```python
#!/usr/bin/env python

from dynaconf import settings
print(settings.TESTVAL)

```
</details>

4. Executing under the following environment

<details>
<summary> Execution </summary>

**invocation with path**
```bash
$ pip install dynaconf==2.0.2

$ cd test2
$ ../test1/pythonscript
test1

$ pip install dynaconf==1.2.1
$ ../test1/pythonscript
test2
```

**invocation via PATH environment**
```bash
$ pip install dynaconf==2.0.2
$ cd test2
$ export PATH=../test1:$PATH
$ pythonscript
test1

$ pip install dynaconf==1.2.1
$ pythonscript
test2
```

</details>

**Expected behavior**
2.0.2 should use the current path for script invocations (1.2.1 versus 2.0.2)

**Environment (please complete the following information):**
 - OS: Linux/Ubuntu18.04
 - Dynaconf Version 2.0.2/pip

rochacbruno:
Hi, in 2.0.x the order of file loading changed, and the expected behavior now is different from the 1.x releases. Because of that breaking change we bumped the version to 2.x.

I was thinking in a way for keeping the compatibility through an environment variable, does that work for you?

something like:

`export DYNACONF_FILE_SEARCH=legacy`

PyUser5:
Hi! Thanks for your reply.

I'm fine with the new version behaving in a different way. I just thought this slipped through. 
(If it is this way I don't need the parameter at all)

The new semantics however may have unintended consequences:
[...]

PR #232, opened 2019-09-13
Fixing self._root_path to fall back to os.getcwd()

kedark3:
I was hitting an error like below:
```---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-c4a32dfb6df1> in <module>
----> 1 cfme_data.management_systems

/var/ci/workspace/kkulkarn-9315/integration_tests/cfme/utils/config_data.py in __getattr__(self, key)
     14 
     15     def __getattr__(self, key):
---> 16         self.config = self.settings[key.upper()]
     17         return self.config
     18 

/var/ci/cfme_venv3/lib64/python3.7/site-packages/dynaconf/utils/functional.py in inner(self, *args)
      8     def inner(self, *args):
      9         if self._wrapped is empty:
---> 10             self._setup()
     11         return func(self._wrapped, *args)
     12 

/var/ci/cfme_venv3/lib64/python3.7/site-packages/dynaconf/base.py in _setup(self)
    124         settings_module = os.environ.get(environment_variable)
    125         self._wrapped = Settings(
--> 126             settings_module=settings_module, **self._kwargs
    127         )
    128         self.logger.debug("Lazy Settings _setup ...")

/var/ci/cfme_venv3/lib64/python3.7/site-packages/dynaconf/base.py in __init__(self, settings_module, **kwargs)
    188         self._defaults = kwargs
    189         self.logger.debug("Initializing Dynaconf (%s)", self._store)
--> 190         self.execute_loaders()
    191 
    192     def __call__(self, *args, **kwargs):

/var/ci/cfme_venv3/lib64/python3.7/site-packages/dynaconf/base.py in execute_loaders(self, env, silent, key, filename)
    855             self.logger.debug("Dynaconf executing: %s", loader.__name__)
    856             loader.load(self, env, silent=silent, key=key)
--> 857         self.load_includes(env, silent=silent, key=key)
    858         self.logger.debug("Loaded Files: %s", deduplicate(self._loaded_files))
    859 

/var/ci/cfme_venv3/lib64/python3.7/site-packages/dynaconf/base.py in load_includes(self, env, silent, key)
    866             from IPython import embed
    867             embed()
--> 868             self.load_file(path=includes, env=env, silent=silent, key=key)
    869             # ensure env vars are the last thing loaded after all includes
    870             last_loader = self.loaders and self.loaders[-1]

/var/ci/cfme_venv3/lib64/python3.7/site-packages/dynaconf/base.py in load_file(self, path, env, silent, key)
    894                     # continue the loop.
    895                     continue
--> 896                 filepath = os.path.join(self._root_path, _filename)
    897                 self.logger.debug("File path is %s", filepath)
    898                 # Handle possible *.globs sorted alphanumeric

/usr/lib64/python3.7/posixpath.py in join(a, *p)
     78     will be discarded.  An empty last part will result in a path that
     79     ends with a separator."""
---> 80     a = os.fspath(a)
     81     sep = _get_sep(a)
     82     path = a

TypeError: expected str, bytes or os.PathLike object, not NoneType

```

To fix this I had to use `os.getcwd()` . I think one could use `ROOT_PATH_FOR_DYNACONF` but I feel `os.getcwd()` is sensible default.

Please review. 

Thanks.

### DIFF
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -892,7 +892,9 @@ class Settings(object):
                     # continue the loop.
                     continue
 
-                filepath = os.path.join(self._root_path, _filename)
+                filepath = os.path.join(
+                    self._root_path or os.getcwd(), _filename
+                )
                 self.logger.debug("File path is %s", filepath)
                 # Handle possible *.globs sorted alphanumeric
                 for path in sorted(glob.glob(filepath)):


rochacbruno:
There is a discussion about it going on on #193 

Before 2.0.0 `os.getcwd` was the default root path, then we faced some problems, some users reported they do not wanted the file search to start on the current path but in the scripts path.

On 2.0.0 we changed to use the path where the program is located to build the search tree and do not use `.` as default.

This is something we need to discuss more to get to a decision.

In meantime I think we can solve this bug by checking for `None` before calling the function.

kedark3:
Thanks Bruno. so you suggest checking `self._root_path, _filename` both against `None`?

rochacbruno:
@kedark3 jsut now saw that this is only for the `load_file` method when called directly, so I think your fix is ok!

Just for curiosity, are you calling `settings.load_file(...)` directly? what is your use case?

kedark3:
@rochacbruno so actually as you may see the traceback in PR description, I was running on Jenkins and my Settings were using LazySettings as follows:
```
 self.settings = LazySettings(
            ENV_FOR_DYNACONF=env,
            INCLUDES_FOR_DYNACONF='{}/*.yaml'.format(<path>)
        )
```

so when I would initialize settings, it was failing there. So Adding `ROOT_PATH_FOR_DYANCONF` solved the problem but I thought it would make sense to fall back on `os.getcwd()` hence proposed the fix. Thanks for quick merge!!

rochacbruno:
A possible way to load different files on demand is

```py
from dynaconf import settings

other_settings = settings.from_env('development', SETTINGS_FILE_FOR_DYNACONF='path/to/otherfile.*')
```

The above will `clone` the current `settings` object in a new instance pointing to the new file provided.

rochacbruno:
@kedark3 nice! thanks for reporting

Issue #237, opened 2019-09-19
[RFC] When trying to load yamls, we would like to load `.local.yaml` files last and auto-merge keys


kedark3:
**Is your feature request related to a problem? Please describe.**
We are trying to replace https://github.com/seandst/yaycl usage with Dynaconf and yaycle has a feature where it lets you have yaml files with 2 types:
1) file.yaml
2) file.local.yaml
When you load the yamls from your conf directory, yaycl loads both files and key-value pairs from `.local.yaml` file would override values from `.yaml` in the AttrDict that is loaded finally 

**Describe the solution you'd like**
I'd like to have similar ability in the Dynaconf

**Describe alternatives you've considered**
Tried using dynaconf_merge and dynaconf_merge_unique options as Bruno suggested but didn't work. 

**Additional context**
This, according to my team lead, is a common practice to have `filename.local.yaml` file that can override values in `filename.yaml` file where `filename` is same.

kedark3:
There are couple of possible solutions:
In the `load_file`(https://github.com/rochacbruno/dynaconf/blob/master/dynaconf/base.py#L872) method :
1) If I modify the sorting order to reverse sort the files as follows:
```for path in sorted(glob.glob(filepath), reverse=True):```
that loads the `.local.yaml` files at last giving us what we need. Although not sure if this is a good solution. 
2) We can, otherwise, pass `INCLUDES_FOR_DYNACONF` as follows:

```INCLUDES_FOR_DYNACONF=['conf/*.yaml', 'conf/*.local.yaml'],```

That will load the `.local.yaml` files after `.yaml` files, but for that we need to update the `if` statement on dynaconf/base.py#L902 to look like:
```if path in already_loaded and '.local.' not in path:```
That way we continue to load the `local.yaml` file properly. This way, `.local.yaml` is loaded last and it properly overrides the values from `.yaml` files. And also merges the dicts. 

3) Or we can read paths, and filter out `.local.` paths as follows so that they are loaded last: 
```
paths = [p for p in sorted(glob.glob(filepath)) if '.local.' not in p]
local_paths = [p for p in sorted(glob.glob(filepath)) if '.local.' in p]
# Handle possible *.globs sorted alphanumeric
for path in paths + local_paths:
    self.logger.debug("Loading %s", path)
    if path in already_loaded:  # pragma: no cover
        self.logger.debug("Skipping %s, already loaded", path)
        continue
    settings_loader(
        obj=self,
        env=env,
[...]

PR #238, opened 2019-09-19
Fixing how filename.local.* files are loaded

kedark3:
Addresses issue discussed in #237


rochacbruno:
Hi @kedark3 sorry the delay,

LGTM :+1: 

But we miss an entrance in documentation, somewhere in https://github.com/rochacbruno/dynaconf/blob/master/docs/guides/usage.md#the-settings-files

I'll merge it and add the documentation entry + a functional test case for this.

Thanks

kedark3:
@rochacbruno thanks. When is the next release :) so that we can get this feature

PR #242, opened 2019-10-08
fix crash on empty settings

paskozdilar:
Dynaconf crashes when used with empty settings.

### Steps to reproduce:

Create a settings.yaml file with the following content:
`default: ~`

Run the following:
```
from dynaconf import settings
settings.reload()
```

It crashes with an `AttributeError: 'NoneType' object has no attribute 'items'`

### Fix

I managed to trace it to a bug in the BaseLoader:
```
data = {}
try:
    data = source_data[env]
except KeyError:
[...]
data = {upperfy(k): v for k, v in data.items()}
```

This works fine when env doesn't exist, or when environment contains values, but "source_data[env]" is `None` when passed an empty settings file. Adding `data = source_data[env] or {}` fixes it and is consistent with the definition of data that was the line before.

### DIFF
--- a/dynaconf/loaders/base.py
+++ b/dynaconf/loaders/base.py
@@ -122,7 +122,7 @@ class BaseLoader(object):
                 env = env.lower()  # lower for better comparison
                 data = {}
                 try:
-                    data = source_data[env]
+                    data = source_data[env] or {}
                 except KeyError:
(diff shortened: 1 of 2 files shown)


paskozdilar:
I added the test cases in the PR.
If anything else is needed, please let me know.

PR #259, opened 2019-11-11
add redis auth

luzihang123:


dynaconf/default_settings.py
```diff
@@ -135,6 +135,7 @@ def reload(*args, **kwargs):
     "host": get("REDIS_HOST_FOR_DYNACONF", "localhost"),
     "port": int(get("REDIS_PORT_FOR_DYNACONF", 6379)),
     "db": int(get("REDIS_DB_FOR_DYNACONF", 0)),
+    "password": str(get("REDIS_AUTH_FOR_DYNACONF", None)),
     "decode_responses": get("REDIS_DECODE_FOR_DYNACONF", True),
 }
 REDIS_FOR_DYNACONF = get("REDIS_FOR_DYNACONF", default_redis)
```
rochacbruno: we cannot send empty passwords do the solution might be different here.

```py 
# Default values for redis configs
default_redis = {
    "host": get("REDIS_HOST_FOR_DYNACONF", "localhost"),
    "port": int(get("REDIS_PORT_FOR_DYNACONF", 6379)),
    "db": int(get("REDIS_DB_FOR_DYNACONF", 0)),
    "decode_responses": get("REDIS_DECODE_FOR_DYNACONF", True),
}
redis_password = get("REDIS_AUTH_FOR_DYNACONF", None)
if redis_password is not None:
    default_redis['password'] = redis_password
REDIS_FOR_DYNACONF = get("REDIS_FOR_DYNACONF", default_redis)
```

luzihang123: You are right. I didn't think it through when password is empty

dynaconf/default_settings.py
```diff
@@ -137,6 +137,10 @@ def reload(*args, **kwargs):
     "db": int(get("REDIS_DB_FOR_DYNACONF", 0)),
     "decode_responses": get("REDIS_DECODE_FOR_DYNACONF", True),
 }
+redis_password = get("REDIS_AUTH_FOR_DYNACONF", None)
+if redis_password:
```
rochacbruno:
```suggestion
if redis_password is not None:
```

so we allow empty strings as password.

dynaconf/default_settings.py
```diff
@@ -137,6 +137,10 @@ def reload(*args, **kwargs):
...
+    default_redis['password'] = redis_password
+REDIS_FOR_DYNACONF = get("REDIS_FOR_DYNACONF", default_redis)
```
rochacbruno:
```suggestion
```

this line is duplicated

luzihang123: Thanks for pointing out that I was careless

PR #292, opened 2020-02-03
Fix #145 allow string formatting using envvars

### LINKED ISSUE #145: RFC: Template Substitutions - dynaconf-render-template plugin
Require: #146 

## Allow template substitutions

There are 2 types of variables that would be useful to have access within dynaconf settings

- Environment Variables `${HOME}`, `${PWD}`
- Dynaconf variables `settings.get('value')`, `settings.ROOT_PATH_FOR_DYNACONF`

When setting a new variable it would be nice to have access to that vars like using special marker `$D{}` or `@D{}` (to be defined).

To avoid overhead, this feature must be enabled via plugin.

```python
pip install dynaconf-template-render
```

```yaml
default:
  databases:
    default:
      name: foo_database
      driver: sqlite3
      file: $D{ROOT_PATH_FOR_DYNACONF}/foo.db
   conf_location: $D{HOME}/.config/foo/  
```

Then after executing the loaders (the bottom of `.execute_loaders` method) there will be a `plugin_hook` which will be called having `obj` on context, then plugin can make changes to `obj` (in place)

rochacbruno:
## Template substitutions 

Dynaconf has 2 tokens to enable string substitutions `@format` and `@jinja`.

### @format token

Dynaconf allows template substitutions for strings values, by using the `@format` token prefix and including placeholders accepted by Python's `str.format` method Dynaconf will call
it lazily upon access time.

The call will be like:

```py
"<YOURVALUE>".format(env=os.environ, this=dynaconf.settings)
```

So in your string you can refer to environment variables via `env` object, and also to variables defined int the settings object itself via `this` reference. It is lazily evaluated on access it will use the final value for a settings regardless the order of load.

Example:

```bash
export PROGRAM_NAME=calculator
```

settings.toml

```toml
[default]
DB_NAME = "mydb.db"

[development]
DB_PATH = "@format {env[HOME]}/{this.current_env}/{env[PROGRAM_NAME]}/{this.DB_NAME}"
```

- `{env[HOME]}` is the same as `os.environ["HOME"]` or `$HOME` in the shell.
- `{this.current_env}` is the same as `settings.current_env`
- `{env[PROGRAM_NAME]}` is the same as `os.environ["PROGRAM_NAME"]` or `$PROGRAM_NAME` in the shell.
- `{this.DB_NAME}` is the same as `settins.DB_NAME` or `settings["DB_NAME"]`

so in your `program`

```py
from dynaconf import settings

settings.DB_PATH == '~/development/calculator/mydb.db'
```

### @jinja token

If `jinja2` package is installed then dynaconf will also allow the use jinja to render string values.

Example:

```bash
export PROGRAM_NAME=calculator
```

settings.toml

```toml
[default]
DB_NAME = "mydb.db"

[development]
DB_PATH = "@jinja {{env.HOME}}/{{this.current_env | lower}}/{{env['PROGRAM_NAME']}}/{{this.DB_NAME}}"
```

so in your `program`

```py
from dynaconf import settings

settings.DB_PATH == '~/development/calculator/mydb.db'
```

The main difference is that Jinja allows some Python expressions to be avaluated such as `{% for, if, while %}` and also supports calling methods and has lots of filters like `| lower`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ test_examples:
 	cd example/;pwd;python compat.py
 	cd example/app;pwd;python app.py
 	cd example/dunder;pwd;python app.py
+	cd example/format;pwd;python app.py
 	cd example/app_with_dotenv;pwd;python app.py
 	cd example/merge_enabled;pwd;python app.py
 	cd example/new_merge;pwd;python app.py
--- a/docs/guides/environment_variables.md
+++ b/docs/guides/environment_variables.md
@@ -54,6 +54,14 @@ DYNACONF_NONE='@none None'
 
 # toml syntax does not allow mixed type arrays so use @json
 DYNACONF_ARRAY='@json [42, 3.14, "hello", true, ["otherarray"], {"foo": "bar"}]'
+
+# Lazily formatted string can access env vars and settings variables.
+
+# using str.format
+DYNACONF_DATABASE_PATH="@format {env[HOME]}/.config/databases/{this.DB_NAME}"
+
+# using jinja2
+DYNACONF_DATABASE_PATH="@jinja {{env.HOME}}/.config/databases/{{this.DB_NAME}}"
 ```
 
 > **NOTE**: Older versions of Dynaconf used the `@casting` prefixes for env vars like `export DYNACONF_INTEGER='@int 123'` still works but this casting is deprecated in favor of using TOML syntax described above. To disable the `@casting` do `export AUTO_CAST_FOR_DYNACONF=false`
--- a/docs/guides/usage.md
+++ b/docs/guides/usage.md
@@ -501,7 +501,85 @@ settings.load_file(path="/path/to/file.toml")  # list or `;/,` separated allowed
 
 > **NOTE**: programmatically loaded file is not persisted, once `env` is changed via `setenv|ugin_env`, or a `reload` or `configure` is invoked it will be cleaned, to persist it needs to go to `INCLUDES_FOR_DYNACONF` variable or you need to load it programmatically again.
 
-## Merging existing values
+## Template substitutions
+
+Dynaconf has 2 tokens to enable string substitutions `@format` and `@jinja`.
+
+### @format token
+
+Dynaconf allows template substitutions for strings values, by using the `@format` token prefix and including placeholders accepted by Python's `str.format` method Dynaconf will call
+it lazily upon access time.
+
+The call will be like:
+
+```py
+"<YOURVALUE>".format(env=os.environ, this=dynaconf.settings)
+```
+
+So in your string you can refer to environment variables via `env` object, and also to variables defined int the settings object itself via `this` reference. It is lazily evaluated on access it will use the final value for a settings regardless the order of load.
+
+Example:
+
+```bash
+export PROGRAM_NAME=calculator
+```
+
+settings.toml
+
+```toml
+[default]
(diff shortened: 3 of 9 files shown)


rochacbruno:
This is also aligned with https://pulp.plan.io/issues/4679#note-3 @pulp issue so @bmbouter @dkliban your review/feedback is also welcome.

rochacbruno:
@gabicavalcante can you please give feedback if this fixes your issue?

you can give a try using:
```bash
pip uninstall dynaconf
pip install https://github.com/rochacbruno/dynaconf/tarball/lazy_format
```

gabicavalcante:
> @gabicavalcante can you please give feedback if this fixes your issue?
> 
> you can give a try using:
> 
> ```shell
> pip uninstall dynaconf
> pip install https://github.com/rochacbruno/dynaconf/tarball/lazy_format
> ```

I tried, but I got an error in this code:
```
In [1]: from dynaconf import settings                                                                                                                                                 
In [2]: settings.get('SQLALCHEMY_DATABASE_URI')     
...
AttributeError: 'str' object has no attribute 'lower()'
``` 

My settings.toml:
```
...
[development]
DB_NAME = "dev.sqlite"
SQLALCHEMY_DATABASE_URI="@format {env[HOME]}/{self.current_env.lower()}/{env[PROGRAM_NAME]}/{self.DB_NAME}"
...
```

rochacbruno:
Hi @gabicavalcante this was my bad, thanks for trying.

Looks like cannot call methods inside the string formatting.

try changing your settings removing the `.lower()` call

```toml
[development]
DB_NAME = "dev.sqlite"
SQLALCHEMY_DATABASE_URI="@format {env[HOME]}/{self.current_env.}/{env[PROGRAM_NAME]}/{self.DB_NAME}"
```

Maybe it is a good idea to use Jinja here?

rochacbruno:
Hi @gabicavalcante I changed the implementation, now it supports 2 kinds of templating `@format` and `@jinja`

Read more in: https://github.com/rochacbruno/dynaconf/pull/292#issue-370529838

rochacbruno:
Also changed `self` to `this` inside template strings.

gabicavalcante:
@rochacbruno that's great, I tested now. I set the `DB_NAME` and `PROGRAM_PATH, and it's working. I tried:
```
SQLALCHEMY_DATABASE_URI="@jinja {{env.HOME}}/{{env['PROGRAM_PATH']}}/{{this.DB_NAME}}"
```

I was thinking about add an jinja extension to use `os.path` functionality. To make something like: 
```
SQLALCHEMY_DATABASE_URI="@jinja {{ this.DB_NAME | abspath }}"
```

rochacbruno:
@gabicavalcante done!

Jinja supports its built-in filters listed in [Builtin Filters Page](http://jinja.palletsprojects.com/en/master/templates/#builtin-filters) and Dynaconf includes aditional filters for `os.path` module: `abspath`. `realpath`, `relpath` and `dirname` and usage is like: `VALUE = "@jinja {{this.FOO | abspath}}"`

rochacbruno:
just added also `basename` seems to be useful

gabicavalcante:
@rochacbruno perfect! I tried using `abspath` just now, and you solve all your problems 😃 
```
DB_NAME = "development.db"
SQLALCHEMY_DATABASE_URI="@jinja {{ this.DB_NAME | abspath }}"
```

PR #307, opened 2020-02-29
Added OSX builds to the Azure Pipeline

### LINKED ISSUE #274: [RFC] Add OS X to CI
Looks like we can have OSX builds

https://devblogs.microsoft.com/devops/azure-pipelines-now-supports-additional-hosted-macos-versions/

We need to add it to our Azure Pipeline

tiagocordeiro:
This close #274


VaultVulp:
Sadly, macOS-based environments don't have a pre-installed docker, thus we have failing steps in our pipeline.

As far as I see, we have a couple of options:
1. Disable tests for Redis and Vault in the OSX environments;
2. Install Docker CLI and Docker daemon before running the tests;
3. Run Redis and Vault in parallel with our test runs, so all pipelines can access them without an installed Docker daemon;

I prefer the third option, but, after reading Azure Pipeline docs, I'm still not sure either we could run a service in parallel with a task or not.

rochacbruno:
@tiagocordeiro @VaultVulp I think we can skip docker tests on O.Sx  the same way we did for windows.

tiagocordeiro:
@rochacbruno @VaultVulp sorry for the mess in the commits

PR #317, opened 2020-03-11
Considering `silent` on import_from_filename

### LINKED ISSUE #310: [bug] POSIX permissions preventing access to settings file fail silently
I put a config file at `/etc/pulp/settings.py` and I indicate to Pulp on this specific system to use it with the envvar `PULP_SETTINGS="/etc/pulp/settings.py"`. Note my user does not have POSIX read permissions to `/etc/pulp/` itself. Pulp configured dynaconf to use `PULP_` as its envvar prefix.

When using `dynaconf list` your settings don't apply and that passes silently. If the user configures a specific file and you cannot read it I expect an error to be raised, or if the file is missing or empty at least a warning.

fao89:
Making errors visible when silent=False
fixes #310

```
➜  dynaconf git:(master) ✗ ls -al | grep fao 
-rw-------.  1 root    root        0 Mar 10 21:39 fao.py
```
```python
In [1]: import errno

In [2]: import io

In [3]: try:
   ...:     io.open('fao.py')
   ...: except IOError as e:
   ...:     print(e.errno)
   ...:     print(e.strerror)
   ...:     
13
Permission denied

In [4]: errno.EPERM
Out[4]: 1

In [5]: errno.EACCES
Out[5]: 13
```

### DIFF
--- a/dynaconf/loaders/py_loader.py
+++ b/dynaconf/loaders/py_loader.py
@@ -112,8 +112,6 @@ def import_from_filename(obj, filename, silent=False):  # pragma: no cover
     if not filename.endswith(".py"):
         filename = "{0}.py".format(filename)
 
-    if filename in default_settings.SETTINGS_FILE_FOR_DYNACONF:
-        silent = True
     mod = types.ModuleType(filename.rstrip(".py"))
     mod.__file__ = filename
     mod._is_error = False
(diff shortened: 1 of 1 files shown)


rochacbruno:
The whole file loader is going to change, thanks for your work on this

PR #338, opened 2020-05-12
Fix #272 allow lowercase keys

### LINKED ISSUE #272: Inconsistent/undefined behavior in when accessing settings with lower-case names
Hey, I'd like to discuss this behavior I just encountered when I was using a Dynaconf Settings object alongside a regular dict, and trying to use the same code for accessing both of them.  Assuming there is a Dynaconf Settings instance named `settings` that has the parameter `PORT` set to 8080, the following behavior can be observed when accessing the parameter:

1. Using key ("bracket") notation:

    ```python
    >>> settings["PORT"]
    8080 
    
    >>> settings["port"]
    8080
    ```

1. Using .get:

    ```python
    >>> settings.get("PORT")
    8080 
    
    >>> settings.get("port")
    8080
    ```

1. Using attribute ("dot") notation:

    ```python
    >>> settings.PORT  
    8080 
    
    >>> settings.port
    …
    AttributeError: 'Settings' object has no attribute 'port'
    ```

The outlier here is obviously the lower-case dot notation, right? But here is where the behavior actually differs from a dictionary:

```python
>>> "PORT" in settings
True

>>> "port" in settings
False
```

Taking into account _all_ returned results above, the last statement should _also_ respond `True`, because it contradicts the "dict-like" ways to access it (`.get` and key notation) where the lower-case variable is indeed "in settings".

As far as I can see the documentation does not mention this, which is why I'd like to propose defining a consistent behavior regarding upper/lower-casing parameters in Dynaconf. IMHO the behavior should be consistent across all access patterns, and I think Dynaconf should internally be case-insensitive and support both lower-case and upper-case parameter access. The casing should only matter on environment variables (always to be upper-case) where lower-case variables don't generally work.

What do you think?

rochacbruno:
Fix #272 allow access of lowercase keys

- `settings.lowercase_key` is allowed
- `settings.dynaconf` is a proxy to internal methods
- `settings.__reserved_attributes` validates key names
- `LazySettings __init__ parameters can receive lower case configs`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,7 @@ test_examples:
 	cd example/toml_with_secrets/;pwd;python program.py
 	cd example/envs;pwd;python app.py
 	cd example/envless_mode;pwd;python app.py
+	cd example/lower_read;pwd;python app.py
 	cd example/custom_loader;pwd;python app.py
 	cd example/get_fresh;pwd;python app.py
 	cd example/includes;pwd;python app.py
--- a/docs/guides/advanced_usage.md
+++ b/docs/guides/advanced_usage.md
@@ -17,7 +17,7 @@ Now you want to use Dynaconf, open that `conf.py` or `conf/__init__.py` and do:
 # coding: utf-8
 from dynaconf import LazySettings
 
-config = LazySettings(ENVVAR_PREFIX_FOR_DYNACONF="MYPROGRAM")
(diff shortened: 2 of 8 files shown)


rochacbruno:
It can be combined with other settings

```py
settings = LazySettings(envless_mode=True, lowercase_read=True)
```

PR #355, opened 2020-06-11
Fix documentation for creating Dynaconf instance

jhesketh:
The lazy loader expects the parameters to follow the same syntax as environment variables.


rochacbruno:
Thanks for your PR, but this change is being introduced in the next release 3.0.0

jhesketh:
Ah, well then the published docs need to be versioned somehow. Only the 'latest' is available on readthedocs but that isn't the version published on pypi.

rochacbruno:
@jhesketh sorry for the inconvenience there were some problem when releasing the docs. Thanks for your issue

PR #357, opened 2020-06-16
Lowecase read allowed by default

rochacbruno:


dynaconf/__init__.py
```diff
@@ -9,11 +9,14 @@
 from dynaconf.validator import ValidationError  # noqa
 from dynaconf.validator import Validator  # noqa
 
-settings = LazySettings()
+"""This global settings is deprecated from 3.0.0+"""
+settings = LazySettings(warn_dynaconf_global_settings=True)
+
```
VaultVulp, line 14:
```suggestion
def __construct_singleton_dynaconf_instance():
    warnings.warn(
        "Usage of `from dynaconf import settings` is now "
        "DEPRECATED in 3.0.0+. You are encouraged to change it to "
        "your own instance e.g: `settings = Dynaconf(*options)`",
        DeprecationWarning,
    )
    return LazySettings()

settings = __construct_singleton_dynaconf_instance()
```

In this case we would not need to change `LazySettings` class.

rochacbruno: I like that, using a FActory/Builder instead of class state, thanks will change it.

VaultVulp: Sorry, my bad. In this case we will always raise a warning. Even if the user created `LazySettings` instance, in an adequate way.

rochacbruno: @VaultVulp yes the idea is nice but it is not going to be lazy.

PR #373, opened 2020-07-16
Add support for vault to auth with iam and token

richardpen:
This PR added the following changes:
* Authenticate vault server with token
* Authenticate vault server with AWS IAM
* Retrieve all the secrets under the vault path if the given path is a directory


dynaconf/loaders/vault_loader.py
```diff
@@ -2,6 +2,7 @@
 # pip install hvac
 from dynaconf.utils import build_env_list
 from dynaconf.utils.parse_conf import parse_conf_data
+import boto3
```
rochacbruno, line 5: boto should be included as a dependency

probably following the same pattern as seen in line 8.

richardpen: Thanks for taking a look @rochacbruno. Also the pipeline failed with the error `ERROR: radon 4.1.0 has requirement colorama==0.4.1, but you'll have colorama 0.4.3 which is incompatible` which seems related to rubik/radon#195. Is there a way to get around that, or this has to be fixed when the radon fix released? I'm pretty new to python, please suggest.

rochacbruno:
we need a rebase and fix for he formatting, locally you can run `make run-pre-commit`

richardpen:
@rochacbruno I have fixed the format and the test, could you take another look, thanks?

richardpen:
@rochacbruno ping for reviewing.

VaultVulp:
Hello, @richardpen !

I'm running the latest vault version: 1.6.0 and using the second version of secrets in it. Right now, I cannot load configuration from my instance.

In this PR you have added a third instance of `data.get("data", {})`, but response from my vault instance have only two levels of data `{"data": {"data": ..., ...}, ...} `, and we already remove them successfully via the `data = data.get("data", {}).get("data", {})`.

To fix this issue, I have to remove following lines from the `vault_loader.py`:
```python
            if obj.VAULT_KV_VERSION_FOR_DYNACONF == 2 and data:
                data = data.get("data", {})
```
but this change can break your use case.

Are you sure that, the third instance of `data.get("data", {})` is required? Maybe you have an example response that have this data structure?

Thanks in advance.

PR #378, opened 2020-07-29
Add auth username and password for redis settings

jiranun:


rochacbruno:
we need a rebase here

PR #387, opened 2020-08-09
Allow testing against local redis server

jayvdb:
This is used for https://build.opensuse.org/request/show/825159 , so that the redis integration can be tests in an isolated environment without access to the internet.

### DIFF
--- a/tests/test_redis.py
+++ b/tests/test_redis.py
@@ -15,14 +15,24 @@ def custom_checker(ip_address, port):
     return True
 
 
-@pytest.fixture(scope="module")
-def docker_redis(docker_services):
-    docker_services.start("redis")
-    public_port = docker_services.wait_for_service(
-        "redis", 6379, check_server=custom_checker
-    )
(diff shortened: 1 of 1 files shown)


tests/test_redis.py
```diff
@@ -15,14 +15,20 @@ def custom_checker(ip_address, port):
...
 
-@pytest.fixture(scope="module")
-def docker_redis(docker_services):
-    docker_services.start("redis")
-    public_port = docker_services.wait_for_service(
-        "redis", 6379, check_server=custom_checker
-    )
-    url = f"http://{docker_services.docker_ip}:{public_port}"
-    return url
+REDIS_URL = os.environ.get('REDIS_URL', None)
```
jayvdb, line 18: suggestion on what to call this variable?  Or alternative ways to achieve this?

(seems I also need to blacken it)

rochacbruno: I would use a more explicit variable name like: `DYNACONF_TEST_REDIS_URL` yeah it is very long but will not get confused with some general REDIS_URL that might exist.

PR #413, opened 2020-09-16
Fix #407 - add proper equality test for CombinedValidator

### LINKED ISSUE #407: [bug] Only one CombinedValidator is registered - subsequent are silently ignored
**Describe the bug**
If validators are added through `settings.validators.register()`, only first CombinedValidator is registered - subsequent are silently skipped.

**Analysis**
The root cause is `Validator.__eq__()` method. `ValidatorList.register()` will go through provided validators and add them, but only if they aren't already on the list (`validator not in self`). `in` will return "`True` if an item of *s* is equal to *x*" ([docs](https://docs.python.org/3.8/library/stdtypes.html#common-sequence-operations)). That logic was added in #256 .

When `Validator.__eq__()` compares different objects, it looks into various `Validator` properties and compares them in pair. If they are all the same, `__eq__()` will assume these are two instances of effectively the same validation rule.

The problem is, `CombinedValidator` doesn't have any of these properties, so two completely different `CombinedValidator` will appear to be the same for `__eq__()` method.

**To Reproduce**
In python shell:
```
>>> from dynaconf import Validator
>>> (Validator("foo.foo") | Validator("foo.bar")) == (Validator("bar.foo") & Validator("bar.bar"))
True
```

This should return False, as these two `CombinedValidator`s have nothing in common.

 - OS: Linux/Fedora32
 - Dynaconf master (6c568d687e29ca5ed9806a74f1f4fb7e4b96be2f), 3.0.1

**Additional context**
I might try working on patch, but I'm not sure about best approach. Perhaps we need type comparison inside `__eq__()` (so comparing AndValidator to OrValidator can return early). But how do we compare two `AndValidator`s? Look into combined validators properties recursively?

mirekdlugosz:
Fix bug reported as #407 by introduction of proper equality test in `CombinedValidator`.

I considered adding new `combined_attributes` property to `Validator`, something that would return tuple of all attributes mentioned in `__eq__` method. Then `CombinedValidator` would override this property and return tuple of two validators `combined_attributes`.

However, the other method didn't allow to provide custom properties values by instantiating `CombinedValidator` directly. In current approach, following does work:

```python
from dynaconf.validator import Validator, AndValidator
a1 = AndValidator(Validator("FOO"), Validator("BAR"), when=Validator("BASE", must_exist=True))
a2 = AndValidator(Validator("FOO"), Validator("BAR"), when=Validator("FOUNDATION", must_exist=True))
assert a1 != a2
```

I rewrote equality tests to use pytest parameters, to extend checking for combined validators that were not checked before. I'm not huge fan of readability of these tests now, so let me know if you would prefer separate test for each input pair.

```
$ pytest -v -k equality tests/test_validators.py 
===================================================================================== test session starts =====================================================================================
platform linux -- Python 3.8.5, pytest-6.0.1, py-1.9.0, pluggy-0.13.1 -- /home/mzalewsk/.virtualenvs/dynaconf/bin/python3
cachedir: .pytest_cache
rootdir: /home/mzalewsk/sources/dynaconf, configfile: pytest.ini
plugins: cov-2.10.1, mock-3.3.1, lovely-pytest-docker-0.1.0
collected 26 items / 17 deselected / 9 selected                                                                                                                                               

tests/test_validators.py::test_equality[this_validator0-another_validator0] PASSED                                                                                                      [ 11%]
tests/test_validators.py::test_equality[this_validator1-another_validator1] PASSED                                                                                                      [ 22%]
tests/test_validators.py::test_equality[this_validator2-another_validator2] PASSED                                                                                                      [ 33%]
tests/test_validators.py::test_inequality[this_validator0-another_validator0] PASSED                                                                                                    [ 44%]
tests/test_validators.py::test_inequality[this_validator1-another_validator1] PASSED                                                                                                    [ 55%]
tests/test_validators.py::test_inequality[this_validator2-another_validator2] PASSED                                                                                                    [ 66%]
tests/test_validators.py::test_inequality[this_validator3-another_validator3] PASSED                                                                                                    [ 77%]
tests/test_validators.py::test_inequality[this_validator4-another_validator4] PASSED                                                                                                    [ 88%]
tests/test_validators.py::test_validator_equality_by_identity PASSED                                                                                                                    [100%]

============================================================================== 9 passed, 17 deselected in 0.02s ===============================================================================
```

### DIFF
--- a/dynaconf/validator.py
+++ b/dynaconf/validator.py
@@ -125,6 +125,9 @@ class Validator(object):
         if self is other:
             return True
 
+        if type(self).__name__ != type(other).__name__:
+            return False
+
         identical_attrs = (
             getattr(self, attr) == getattr(other, attr)
             for attr in (
@@ -229,6 +232,24 @@ class CombinedValidator(Validator):
         """Takes 2 validators and combines the validation"""
         self.validators = (validator_a, validator_b)
         super().__init__(*args, **kwargs)
+        self.names = self.names or tuple(
+            validator.names for validator in self.validators
+        )
+        self.must_exist = self.must_exist or tuple(
+            validator.must_exist for validator in self.validators
+        )
+        self.when = self.when or tuple(
+            validator.when for validator in self.validators
+        )
+        self.condition = self.condition or tuple(
+            validator.condition for validator in self.validators
+        )
+        self.operations = self.operations or tuple(
+            validator.operations for validator in self.validators
+        )
+        self.envs = self.envs or tuple(
+            validator.envs for validator in self.validators
+        )
 
     def validate(self, settings):  # pragma: no cover
         raise NotImplementedError(
--- a/dynaconf/validator.py
+++ b/dynaconf/validator.py
@@ -5,6 +5,16 @@ from dynaconf import validator_conditions  # noqa
 from dynaconf.utils.functional import empty
 
 
+EQUALITY_ATTRS = (
+    "names",
+    "must_exist",
+    "when",
+    "condition",
+    "operations",
+    "envs",
+)
+
+
 class ValidationError(Exception):
     pass
 
@@ -130,14 +140,7 @@ class Validator(object):
 
         identical_attrs = (
             getattr(self, attr) == getattr(other, attr)
-            for attr in (
-                "names",
-                "must_exist",
-                "when",
-                "condition",
-                "operations",
-                "envs",
-            )
+            for attr in EQUALITY_ATTRS
         )
         if all(identical_attrs):
             return True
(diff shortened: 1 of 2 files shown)


dynaconf/validator.py
```diff
@@ -229,6 +232,24 @@ def __init__(self, validator_a, validator_b, *args, **kwargs):
...
+        )
+        self.condition = self.condition or tuple(
+            validator.condition for validator in self.validators
+        )
+        self.operations = self.operations or tuple(
+            validator.operations for validator in self.validators
+        )
+        self.envs = self.envs or tuple(
+            validator.envs for validator in self.validators
+        )
```
rochacbruno, line 252: what if on the top of the file we define a constant

```py
EQUALITY_ATTRS = ("names", "must_exist", "when", "condition", "operations", "envs")
```

then we can use that on line [133](https://github.com/rochacbruno/dynaconf/pull/413/files#diff-2f5568fbbe07143ea6573f6f64c74bbdR133)

and then here as most of this code is pretty similar we also reuse that attrs?

```suggestion
        for attr in EQUALITY_ATTRS:
            if getattr(self, attr, None) is None:
                setattr(
                    self, 
                    attr,
                    tuple(getattr(validator, attr) for validator in self.validators)
                ) 
```

It is not very readable, but if we add more attrs in the `EQUALITY_ATTRS` list in the future it will be dynamically covered.

mirekdlugosz: Added in 72a2b76c515271d6eaf2f52d38b9f692260d473f . Should I squash?

dynaconf/validator.py
```diff
@@ -229,6 +235,12 @@ def __init__(self, validator_a, validator_b, *args, **kwargs):
         """Takes 2 validators and combines the validation"""
         self.validators = (validator_a, validator_b)
         super().__init__(*args, **kwargs)
+        for attr in EQUALITY_ATTRS:
+            if not getattr(self, attr, None):
```
mirekdlugosz, line 239: Due to `*names` in `Validator` constructor, `self.names` will always be defined (and default to empty tuple). That's why we can't use `if getattr(self, attr, None) is None` here - it would never call `setattr` for `names`.

PR #444, opened 2020-10-07
Fix #443 object merge with same value on same level keys

### LINKED ISSUE #443: [bug] Setting values being stripped out after merge
**Describe the bug**
I'm seeing where values are being stripped out of settings. This is on dynaconf 3.1.1 with Python 3.7.7.

**To Reproduce**
Run this code:

```python
from dynaconf.utils.boxing import DynaBox
from dynaconf.utils import object_merge

existing_data = DynaBox({"DATABASES": {'default': {'ENGINE': 'django.db.backends.postgresql_psycopg2', 'HOST': 'localhost', 'PASSWORD': 'pulp', 'NAME': 'pulp', 'USER': 'pulp'}}})
new_data = DynaBox({'DATABASES': {'default': {'USER': 'postgres'}}})
split_keys = ['DATABASES', 'default', 'USER']
object_merge(old=existing_data, new=new_data, full_path=split_keys)
```

The output:

```
<Box: {'DATABASES': {'default': {'USER': 'postgres', 'ENGINE': 'django.db.backends.postgresql_psycopg2', 'HOST': 'localhost'}}}>
```

Notice that NAME and PASSWORD keys have been dropped.

**Expected behavior**
I would expect the final settings dict to be:

```
{"DATABASES": {'default': {'ENGINE': 'django.db.backends.postgresql_psycopg2', 'HOST': 'localhost', 'PASSWORD': 'pulp', 'NAME': 'pulp', 'USER': 'postgres'}}}
```

rochacbruno:
This solution is a temporary solution as it solves current
problem, but there is still the case for `None` values.

The best solution for this case would be wrapping all the values
on assignment and give it a full path signature to compare.

Fix #443

### DIFF
--- a/dynaconf/utils/__init__.py
+++ b/dynaconf/utils/__init__.py
@@ -42,7 +42,11 @@ def object_merge(old, new, unique=False, full_path=None):
 
         for key, value in old.items():
 
-            if existing_value is not None and existing_value is value:
+            if (
+                existing_value is not None
+                and key == full_path[-1]
+                and existing_value is value
+            ):
                 continue
 
             if key not in new:
--- a/Makefile
+++ b/Makefile
@@ -105,6 +105,7 @@ test_examples:
 	cd example/issues/404_dup_validator_message;pwd;python app.py
 	cd example/issues/434_setenv;pwd;python app.py --config development dynaconf
 	cd example/issues/430_same_name;pwd;python app.py
+	cd example/issues/443_object_merge;pwd;python app.py
 
 test_vault:
 	# @cd example/vault;pwd;python write.py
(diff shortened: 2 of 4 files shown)


dynaconf/utils/__init__.py
```diff
@@ -42,7 +42,11 @@ def object_merge(old, new, unique=False, full_path=None):
 
         for key, value in old.items():
 
-            if existing_value is not None and existing_value is value:
+            if (
+                existing_value is not None
+                and key == full_path[-1]
+                and existing_value is value
```
rochacbruno, line 48: This line is still `buggy` as this comparison cannot handle `None` types.

To improve this we need  to create a way to have a unique signature for each value.

daviddavis:
dance_with_cat

PR #502, opened 2021-01-13
Add __contains__ to Dynaconf

Mdslino:


dynaconf/base.py
```diff
@@ -105,6 +105,9 @@ def __init__(self, **kwargs):
         self._kwargs = kwargs
         super(LazySettings, self).__init__()
 
+    def __contains__(self, item):
+        return hasattr(self, item)
```
rochacbruno, line 109: dynaconf stores data in 2 places.

1. self._store which is a dictionary
2. as instance attributes 

I am not sure if simply calling `hasattr` will use `__getattr__` which already makes the lookup in both places.

Also we might need to add some unit tests and a functional test in /example/ folder

Mdslino: Ok, i get it. I'll test this scenario and write some tests to validate this.

rochacbruno (approved):

PR #532, opened 2021-03-02
Fix #504 implement math and postfix evaluator tokens

### LINKED ISSUE #504: [docs] add docs for jinja Formula/math in numeric fields
Hey all,
I want to raise the discussion as I didn't find a documented solution and the obvious ways do not work.

Is it possible to pass formulas as configuration values? A typical example would be `duration_seconds = 24*60*60` or `factor = 1/60`. Yes, one could write it out but the mathematical expression often holds additional value or is more accurate (second example).

I read about "computed values" but wasn't able to make that work for the example.

Thanks and all the best

rochacbruno:
Fix #504

```toml
duration = "@math 60 * 24"
duration = "@postfix 60 24 *"
```

TODO:

- [ ] Add more tests
- [ ] make it lazy
- [ ] make it work with `format` e.g: `@math {this.FACTOR} * 30`

### DIFF
--- a/dynaconf/utils/math_evaluator.py
+++ b/dynaconf/utils/math_evaluator.py
@@ -1,0 +1,131 @@
+import math
+import operator
+
+_add, _sub, _mul = operator.add, operator.sub, operator.mul
+_truediv, _pow, _sqrt = operator.truediv, operator.pow, math.sqrt
+_sin, _cos, _tan, _radians = math.sin, math.cos, math.tan, math.radians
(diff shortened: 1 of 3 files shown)


rochacbruno:
we already have jinja support, probably we don't want this PR

```toml
duration = "@jinja {{ 24 * 60 }}"
```

rochacbruno:
https://jinja.palletsprojects.com/en/2.11.x/templates/#math

PR #541, opened 2021-03-08
HOTFIX for 462 related issue, `default` on .get should be parsed as Box objects

rochacbruno:
In order to keep the same method api, default values should be parsed
and converted to Boxed objects.

https://github.com/rochacbruno/dynaconf/issues/462


example/issues/449_django_lazy_path/foo/settings.py
```diff
@@ -46,3 +45,21 @@
 assert settings.TEMPLATES[0]["DIRS"] == [
     f"{BASE_DIR}/templates"
 ], settings.TEMPLATES[0]["DIRS"]
+
+ALLOWED_HOSTS = settings.get(
+    "ALLOWED_HOSTS", ["::1", "127.0.0.1", "localhost"]
+)
```
dmsimard, line 51: :+1:

PR #549, opened 2021-03-10
Add the ability for selective validation

### LINKED ISSUE #508: [RFC] Allow for validation of only specific nested sections
There is currently no way to specify what section(s) of a settings object are to be validated during a validate call. This can cause issues when validators are added, but the values they are validating are not.

Example: I have two larger sub-sections of a settings object, where the values for the second have been changed to be invalid. I do not want the second section's validators to trigger when I'm validating the first as that raises a validation error.
```python
# section 1
settings.SECTION1

#section 2
settings.SECTION2

settings.SECTION2.some_key = "bad value"

settings.validators.validate()  # raises a validation error
```

Ideally, I would like to be able to specify which sections of the settings object should be subject to validation when calling validate.

Example:
```python
# only validate SECTION1 settings, so SECTION2 validation errors don't interfere
settings.validators.validate(settings.SECTION1) 
```

Optional: Add `only` and/or `exclude` arguments to the validate method.

Example:
```python
settings.validators.validate(only=[settings.SECTION1, settings.SECTION3])

settings.validators.validate(exclude=settings.SECTION2)
```

JacobCallahan:
This change introduces the ability to control which sections of a
settings object are subject to validation.
This is controlled primarily by two mechanisms.
1: When creating a settings object, new arguments `validate_only`
   and `validate_exclude` have been added which receive a list of settings
   paths.
2: When manually calling validate, new arguments `only` and `exclude`
   have been added.
All of these allow for either a string or list of strings representing
settings paths. For example:
    `settings.validators.validate(only=["settings.something",
    "settings.another"])`

    settings = Dynaconf(..., validate_exclude="settings.bad")

Fixes #508

### DIFF
--- a/dynaconf/validator.py
+++ b/dynaconf/validator.py
@@ -2,6 +2,7 @@ from itertools import chain
 from types import MappingProxyType
 
 from dynaconf import validator_conditions  # noqa
+from dynaconf.utils import ensure_a_list
 from dynaconf.utils.functional import empty
 
 
@@ -147,8 +148,16 @@ class Validator:
 
         return False
 
-    def validate(self, settings):
+    def validate(self, settings, only=None, exclude=None):
         """Raise ValidationError if invalid"""
+        # If only or exclude are not set, this value always passes startswith
+        only = ensure_a_list(only or [""])
+        if only and not isinstance(only[0], str):
+            raise ValueError("'only' must be a string or list of strings.")
+
+        exclude = ensure_a_list(exclude)
+        if exclude and not isinstance(exclude[0], str):
+            raise ValueError("'exclude' must be a string or list of strings.")
 
         if self.envs is None:
             self.envs = [settings.current_env]
@@ -159,7 +168,7 @@ class Validator:
                 if self.when.envs is None:
                     self.when.envs = self.envs
 
-                self.when.validate(settings)
+                self.when.validate(settings, only=only, exclude=exclude)
             except ValidationError:
                 # if when is invalid, return canceling validation flow
                 return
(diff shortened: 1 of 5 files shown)


JacobCallahan:
todo:
- [ ] fix logic behind failing tests
- [ ] update docs
- [ ] potentially add more tests

example/validators/with_python/conf.py
```diff
@@ -45,6 +45,8 @@
     Validator(
         "A_DICT.NESTED_1.NESTED_2.NESTED_3.NESTED_4", must_exist=True, eq=1
     ),
+    Validator("A_DICT.NESTED_1.NOT.YET.LOADED", must_exist=True, eq=1),
 )
 
-settings.validators.validate()
+# Validate settings except those that we don't want to validate yet
+settings.validators.validate(validate_exclude="A_DICT.NESTED_1.NOT")
```
rochacbruno, line 52: I guess here it is only `exclude=` and not `validate_exclude` that is why CI is failing

dynaconf/validator.py
```diff
@@ -147,8 +147,23 @@ def __eq__(self, other):
 
         return False
 
-    def validate(self, settings):
+    def validate(self, settings, only=None, exclude=None):
         """Raise ValidationError if invalid"""
+        # If only or exclude are not set, this value always passes startswith
+        only = only or [""]
+        exclude = exclude or []
+        if not isinstance(only, list):
```
rochacbruno, line 155: https://github.com/rochacbruno/dynaconf/blob/master/dynaconf/utils/__init__.py#L241

```py
from dynaconf.utils import ensure_a_list
only = ensure_a_list(only) 
```

dynaconf/validator.py
```diff
@@ -169,15 +184,25 @@ def validate(self, settings):
...
-    def _validate_items(self, settings, env=None):
+    def _validate_items(self, settings, env=None, only=None, exclude=None):
         env = env or settings.current_env
         for name in self.names:
+            # Skip if only is set and name isn't in the only list
+            if not any(name.startswith(sub) for sub in only):
+                continue
+            # Skip if exclude is set and name is in the exclude list
+            if any(name.startswith(sub) for sub in exclude):
+                continue
```
rochacbruno, line 205: We might add to the docs (if not there yet) that it works by matching the first letters.
because the case where the settings has a `FOO` and a `FOOBAR` if `only|exclude` is set to `["FOO"]` both are considered right?

example/validators/with_python/conf.py
JacobCallahan: ahh good catch!!

dynaconf/validator.py
JacobCallahan: Do you have a preference for what to do if someone passes in an incorrect data type?
say someone tries to pass in `.validate(only=settings.some.key)`, that isn't a string or list of strings.
Would you prefer an exception be raised? If so, any preference on exception class?

dynaconf/validator.py
JacobCallahan: That is correct. I can make that more transparent in the docs.

PR #564, opened 2021-04-04
using hvac kv variants in vault loader to get config

michael-nextpart:
Still using deprecated functions for list_secrets and create_or_update_secret. Following up in subsequent PR on that. 😏


rochacbruno:
We need to rebase it and fix the deprecation as in #624

PR #622, opened 2021-07-23
WIP - feat: adding settings file prefix option

zzZIMAWAKE:
Related to: https://github.com/rochacbruno/dynaconf/issues/621

This adds a `settings_file_prefix` option for toml, json, ini and yaml files which works similarly to `envvar_prefix`, however it defaults to `None` and does not filter out any keys if it is not set.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -407,6 +407,17 @@ Read more on [settings_files](/settings_files/)
 
 ---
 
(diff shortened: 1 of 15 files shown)


zzZIMAWAKE:
Closing until I sort this out, made a mistake with not testing examples and need to figure something out prior to opening this, sorry.

Issue #621, opened 2021-07-22
[bug] envvar_prefix is not respected if using a loader other than env


zzZIMAWAKE:
**Describe the bug**
Apologies if this is not considered a bug but I found this behaviour slightly unexpected but perhaps I am misusing things.

If you load your settings from a file (as an example I'll make use of the ini loader) and you have `ENV_PREFIX` set, the settings that are loaded via the ini loader will not strip the prefix in the same way they would were they loaded via the environment.

I would expect all the loaders to behave in the same way stripping the prefix.

**To Reproduce**

1. Create a file: `touch dynaconf_settings.ini`
2. Add the following contents:
```
TESTING_my_env_var=123
what_what=123
wow_wow=123
TESTING_huh=123
```
3. `pip install "dynaconf[ini]"`
4. Set an env var pointing to the settings file - `export SETTINGS_FILE_FOR_DYNACONF=dynaconf_settings.ini`
5. Open up a shell `python3`
6. Run the following:
```
from dynaconf import Dynaconf
settings = Dynaconf(envvar_prefix="TESTING")

settings.my_env_var
<Error>
settings.TESTING_my_env_var=123
<returns '123'>
```

For clarity running the following:
`export TESTING_my_env_var=124`
`python3`
```
from dynaconf import Dynaconf
settings = Dynaconf(envvar_prefix="TESTING")

settings.my_env_var
<returns '124'>
```

as I would expect

**Expected behavior**
`settings.my_env_var` would store the value 123, as it would if this was loaded from an env var

**Environment (please complete the following information):**
 - OS: OSX
 - Dynaconf Version: 3.1.4
 - Frameworks in use: None

**Additional context**

Sorry if this is considered expected behaviour, its thrown a spanner in the works for what I had planned and curious whether this was intended. I'm planning to use dynaconf within a package that will allow users to set up configuration in various ways (env vars / ini files / toml files) and anticipated that they could call the variables the same in any of these cases e.g. `TESTING_my_env_var` in env / ini file / toml files and I'd still be able to reference it within the package as `settings.my_env_var`

If this is indeed a bug and you would like this behaviour to work as described above I'm happy to try and find some time to submit a pull request.

Thanks!

rochacbruno:
@zzZIMAWAKE the env_prefix is meant to be used only for environment variable loading, we don't have a similar feature for other sources, is that useful? what is the use case? do you want to propose this as a new feature?

If you want multiple environments in a setting file you can use

```py
[...]

PR #625, opened 2021-07-23
feat: add filter strategy with PrefixFilter

zzZIMAWAKE:
Related to: #621

This adds a settings_file_prefix option for toml, json, ini and yaml files which works similarly to envvar_prefix, however it defaults to None and does not filter out any keys if it is not set.

If envs are in use it will do the filtering within each env's variables, an example is shown below

```
[default]
PREFIX_EXTRA = "this will supercede if others are set without prefix"

[development]
PREFIX_HOST = "https://example.com/development"
PREFIX_PORT = "8080"
COMMENT = "this will not show when prefix is used"
EXTRA = "will not show"

[production]
PREFIX_HOST = "https://example.com/production"
PORT = "443"
EXTRA = "will not show"
```

When these are loaded up using the `settings_file_prefix="PREFIX"` option you'll get the following:

```
>>> settings.HOST
'https://example.com/development'
>>> settings.PORT
'8080'
>>> settings.COMMENT
AttributeError: 'Settings' object has no attribute 'COMMENT'
```

When switching to production env:
```
>>> with settings.using_env("production"):
...     settings.HOST
...     settings.EXTRA
...     settings.PORT

'https://example.com/production'
'this will supercede if others are set without prefix'
AttributeError: Attribute PORT was deleted, or belongs to different env
```

I think this behaves as you'd expect, looking forward to some feedback as I had to fumble about a bit to learn how everything fits together so no doubt I've done some less than ideal things

Envless load when environments exist in the file will not work in this way as we'd not be able to detect that `settings.PRODUCTION` is an env to filter the prefix from children. I think this is probably expected if loading envless while envs are included in the file - but if it was possible to check if `ENVIRONMENTS_FOR_DYNACONF` is False *and* detect that the top level key is an environment (`[production]`) then it may be possible, but I could not see an obvious way to detect that `production` is an environment

### DIFF
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -35,7 +35,7 @@ from dynaconf.vendor.box.box_list import BoxList
 
 
 class LazySettings(LazyObject):
-    """Loads settings lazily from multple sources::
+    """Loads settings lazily from multiple sources::
 
         settings = Dynaconf(
             settings_files=["settings.toml"],  # path/glob
@@ -203,6 +203,7 @@ class Settings:
         self._not_installed_warnings = []
         self._validate_only = kwargs.pop("validate_only", None)
         self._validate_exclude = kwargs.pop("validate_exclude", None)
+        self._settings_file_prefix = kwargs.get("settings_file_prefix", None)
 
         self.validators = ValidatorList(
             self, validators=kwargs.pop("validators", None)
@@ -537,6 +538,10 @@ class Settings:
             for key in UPPER_DEFAULT_SETTINGS
             if key not in RENAMED_VARS
         }
+
+        if self._settings_file_prefix:
+            # We need to retain the prefix when switching environments
+            new_data["settings_file_prefix"] = self._settings_file_prefix
 
         # This is here for backwards compatibility
         # To be removed on 4.x.x
@@ -1180,5 +1185,6 @@ RESERVED_ATTRS = (
         "validators",
         "_validate_only",
         "_validate_exclude",
+        "_settings_file_prefix",
     ]
 )
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -200,10 +200,10 @@ class Settings:
         self._defaults = DynaBox(box_settings=self)
         self.environ = os.environ
         self.SETTINGS_MODULE = None
+        self.settings_file_prefix = kwargs.get("settings_file_prefix", None)
         self._not_installed_warnings = []
         self._validate_only = kwargs.pop("validate_only", None)
         self._validate_exclude = kwargs.pop("validate_exclude", None)
-        self._settings_file_prefix = kwargs.get("settings_file_prefix", None)
 
         self.validators = ValidatorList(
             self, validators=kwargs.pop("validators", None)
@@ -539,9 +539,9 @@ class Settings:
             if key not in RENAMED_VARS
         }
 
-        if self._settings_file_prefix:
+        if self.settings_file_prefix:
             # We need to retain the prefix when switching environments
-            new_data["settings_file_prefix"] = self._settings_file_prefix
+            new_data["settings_file_prefix"] = self.settings_file_prefix
 
         # This is here for backwards compatibility
         # To be removed on 4.x.x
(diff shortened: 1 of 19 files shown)


zzZIMAWAKE:
I've got to look in to the issue I'm seeing with one of the examples and recursion depth - will take a look at that tomorrow most likely

Will also add the code coverage that I'm missing, happy to receive any feedback now as well if you fancy it :)

zzZIMAWAKE:
Not sure what is happening with the recursive limit being reached for the example: 482_layered_format - it does it for me locally on both master & this branch, have I genuinely broken something?

zzZIMAWAKE:
@rochacbruno This is ready for review, the some parts of the pipeline are taking a while as they are queued but everything seems to pass successfully now.

rochacbruno:
I like this feature, but now that I was able to fully understand its purpose I have some comments:

1. Naming

 `settings_file_prefix` as the name for argument can be confusing, in a first glance it may look like the prefix for a file name instead of variables inside the file.

2. Make it more composable?


Imagine if on the user standing point it can be this:

`dynaconf/strategies/filtering.py`  **new module**
```py

class PrefixFilter:

    def __init__(self, prefix):
        if not isinstance(prefix, str):
            raise TypeError("`SETTINGS_FILE_PREFIX` must be str")
        prefix = "{}_".format(upperfy(prefix))

    def __call__(self, data, settings):
        """Filter incoming data by prefix"""
        prefix = self.prefix
        return {
                upperfy(
                    k[len(prefix):]
                ): v for k, v in data.items()
                if upperfy(k[:len(prefix)]) == prefix
         }

```

on `dynaconf/loaders/base.py` instead of passing `prefix` around as argument we can have the `obj` to hold the `filter_strategy` as an instance of that class.


```py

from dynaconf import Dynaconf
from dynaconf.strategies.filtering import PrefixFilter

settings = Dynaconf(
    settings_file="settings.toml",
    filter_strategy= PrefixFilter("myprefix")
)

```

That `self.filter_strategy` will be a callable that can be used on https://github.com/rochacbruno/dynaconf/pull/625/files#diff-4d2335a3a5b1b1e4cf1f0d9490c21ed91df851ce973b420729a385d372906eb1R167

@zzZIMAWAKE what do you think?

It can accommodate your use case and allow users to be able to customize the filtering with custom classes or any callable that takes `data, settings` arguments.

rochacbruno:
@zzZIMAWAKE I have been thinking in a way to let users to alternatively pass `environments=["prod", "dev", "qa"]` a list of environments explicitly and then "default" and "global" would be automatically part of the environment sets.

It would still accept `True` to load environments arbitrarily but also allow user to specify it. (maybe this is something to be tackled in a different issue/PR)

zzZIMAWAKE:
Would that also be allowed to be passed when `environments=False`? What would be the behaviour in that case? The issue I'm having is that if someone tried to use:

```
from dynaconf import Dynaconf
from dynaconf.strategies.filtering import PrefixFilter

settings = Dynaconf(
    settings_file="settings.toml",
    environments=False
    filter_strategy= PrefixFilter("myprefix")
)
```
with a file that contained:

```
[production]
PREFIX_VAR = TEST
```
It will return blank as it will filter out `production` as it doesn't begin with the prefix so `settings.PRODUCTION.VAR` will not exist as `settings.PRODUCTION` will not even exist. Perhaps this is an edge case that is okay with the new strategy as the user could set up their own filtering strategy.

@rochacbruno Have updated to make use of `filtering_strategy` if you'd like to take a look :)

dynaconf/base.py
```diff
@@ -538,6 +539,10 @@ def from_env(self, env="", keep=False, **kwargs):
             if key not in RENAMED_VARS
         }
 
+        if self.filter_strategy:
+            # Retain the filtering strategy when switching environments
+            new_data["filter_strategy"] = self.filter_strategy
```
zzZIMAWAKE, line 544: Is this the correct way to retain this when `.using_env` is used? How are other settings retained? (`_validate_only / _validate_exclude` etc, or do they not need to be kept when switching envs?)

rochacbruno:
LGTM  at a forst look @zzZIMAWAKE this is on my list to take deeper a look later today or tomorrow, I will reply here. Thanks again.

dynaconf/base.py
rochacbruno: validation happens only on first instantiation

PR #629, opened 2021-07-26
Add a warning if `--env` is passed to `init`

RobertRosca:
Updated the CLI Click help strings, added in a warning print if `--env/-e` is passed to `init` to mention that it is a deprecated option an has no effect, also added a condition to raise an exception if `dynaconf -i ... init` is called as using instance with the init subcommand doesn't work and the standard exception may not be too clear.

Also updated `docs/cli.md` to be in sync with new help output from the subcommands, and to reflect the deprecation of `dynaconf init --env ...`.

Closes https://github.com/rochacbruno/dynaconf/issues/589


RobertRosca:
Thanks for the merge :smile:

PR #632, opened 2021-08-03
when

rochacbruno:


rochacbruno:
Alternative is https://github.com/rochacbruno/dynaconf/issues/654

wgordon17:
@rochacbruno Can you describe a bit more about what you mean with `going to use jinja.env.compile_expression`? Some kind of example or doc would be awesome!

PR #729, opened 2022-03-29
Use default value when settings is blank

jyejare:
Problem Statement:
-----------------------
Currently, when the value for a setting is blank in YAML, the dynaconf settings object is returning the None value instead of what's set as a default parameter.

The Validator example - `Validator('azurerm.client_secret', default='myAzurePass'),` is returning None when client_secret is set blank in azure.yaml

Expected:
-----------
When the Settings is set to blank in the YAML file then it must use the default value.

The Validator example - `Validator('azurerm.client_secret', default='myAzurePass'),` should return 'myAzurePass' when client_secret is set blank in azure.yaml

Solution:
----------
The `setdefault` method in the dynaconf settings class is updated to accept the value as blank to return the value from default parameter.

### DIFF
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -319,7 +319,7 @@ class Settings:
     def setdefault(self, item, default):
         """Returns value if exists or set it as the given default"""
         value = self.get(item, empty)
-        if value is empty and default is not empty:
+        if (not value or value is empty) and default is not empty:
             self.set(
                 item,
                 default,


dynaconf/base.py
```diff
@@ -319,7 +319,7 @@ def values(self):
     def setdefault(self, item, default):
         """Returns value if exists or set it as the given default"""
         value = self.get(item, empty)
-        if value is empty and default is not empty:
+        if (not value or value is empty) and default is not empty:
```
mshriver, line 322: @rochacbruno @jyejare It looks like this has broken integration with environment variables.

If I have dynaconf 3.1.8 installed, and the combination of:

1. settings.yaml field key (nested one level from top map) is present, value is empty
2. validator with default=None for given field key
3. envvar set for nested field key (DYNACONF_something__field_key)
4. settings object configuration that looks like: https://github.com/SatelliteQE/robottelo/blob/master/robottelo/config/__init__.py#L18 

My envvar value is no longer set in the resulting box instance, it takes the validator default of `None`.

I confirmed its this line in particular by modifying the file in site-packages with 3.1.8 installed.

jyejare: @mshriver @rochacbruno :
1. The dynaconf should have the unit test to test this or env var should be prioritise over default if its None.
2. I wonder why `None` explicitly being set as a settings default, it should not be as by default it should be Empty from dynaconf.
3. This could be fixed easily by updating the condition to:
if (not value or value is empty) and default not in  [empty, None]:

rochacbruno: @jyejare can you add these notes to #741 and then if possible open the PR + adding the needed unit tests?

jyejare: @mshriver @rochacbruno Tried reproducing locally and it's not reproducible with 3.1.8! Commented my observations on issue !

PR #759, opened 2022-06-08
Add pyupgrade hook

johnnv1:
As describe in #757, [this hook](https://github.com/asottile/pyupgrade) will automatically upgrade syntax for newer versions of the language.

I also run `pre-commit autoupdate` to updates the hooks (maybe add [pre-coomit ci](https://pre-commit.ci/) to help on this?)

### DIFF
--- a/dynaconf/utils/__init__.py
+++ b/dynaconf/utils/__init__.py
@@ -5,13 +5,8 @@ import warnings
 from collections import defaultdict
 from json import JSONDecoder
 from typing import Any
-from typing import Dict
 from typing import Iterator
-from typing import List
-from typing import Optional
-from typing import Tuple
 from typing import TYPE_CHECKING
-from typing import Union
 
 
 if TYPE_CHECKING:  # pragma: no cover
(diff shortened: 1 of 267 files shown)


dynaconf/utils/__init__.py
```diff
@@ -86,8 +81,8 @@ def object_merge(
 
 
 def recursive_get(
-    obj: Union[DynaBox, Dict[str, int], Dict[str, Union[str, int]]],
-    names: Optional[List[str]],
+    obj: DynaBox | dict[str, int] | dict[str, str | int],
+    names: list[str] | None,
```
rochacbruno, line 85: Same on python 3.8

```
>>> x: str | int
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'type' and 'type'

```

johnnv1:
What do you think to add this to ensure these type annotations on all files?

```yml
hooks:
  - id: reorder-python-imports
    args: [--py38-plus, --add-import, 'from __future__ import annotations']
 ```

Issue #374, opened 2020-07-27
Dynaconf can't find settings file when using pytest


trallnag:
Where do I put settings files / How do I configure Dynaconf so it will work with pytest? Currently I get the following error:

```
AttributeError: 'Settings' object has no attribute 'LOGGING'
```

It seems like Dynaconf is looking for the files in the tree the python command was invoked from and not relative the module itself.

Here is how my project structure (simplified) looks like:

```
.
├── ...
├── project
│   ├── app.py
│   ├── conf.py
│   ├── __init__.py
│   ├── settings.toml
└── tests
    ├── conftest.py    
    └── test_app.py
```

My `conf.py`:

```python
from dynaconf import Dynaconf

settings = Dynaconf(
    envvar_prefix="DYNACONF", settings_files=["settings.toml", ".secrets.toml"],
)
```

Thanks in advance

trallnag:
Oh well I found a solution.

```python
import os

from dynaconf import Dynaconf

current_directory = os.path.dirname(os.path.realpath(__file__))

settings = Dynaconf(
    envvar_prefix="DYNACONF", settings_files=[f"{current_directory}/settings.toml", f"{current_directory}/.secrets.toml"],
)
```

I will keep this issue open for a bit, maybe I can get feedback if this is an appropriate solution or not

MarcoAlejandro:
It worked for me. Thanks @trallnag

LeafwaysWorks:
worked thanks

rochacbruno:
You can also do

```py
import os

from dynaconf import Dynaconf

current_directory = os.path.dirname(os.path.realpath(__file__))

settings = Dynaconf(
    root_path=current_directory,
    envvar_prefix="DYNACONF", 
    settings_files=["settings.toml", ".secrets.toml"],
)

```

vbrinza:
Does it mean we need a fix in the documentation so the others who start working with it do not have to spend few ours figuring it out as I did?

rochacbruno:
@vbrinza yes that would be great to add that to the docs https://www.dynaconf.com/advanced/#testing

mababio:
Ah thanks for the solution!! ... i was 2 hours into debugging

vbrinza:
@rochacbruno please take a look at my PR https://github.com/dynaconf/dynaconf/pull/819

sayurionishi:
Thank you very much!

PR #819, opened 2022-10-09
add document instructions on config files path

vbrinza:
As agreed in this thread https://github.com/dynaconf/dynaconf/issues/374#issuecomment-1265467769, it would be helpful for others to find the correct instructions.


rochacbruno:
#818 fixed this before, thanks

PR #828, opened 2022-11-04
feat: Add hookable functions

rochacbruno:
Allows to add AFTER and BEFORE hooks to some dynaconf methods

```python
from dynaconf import Dynaconf

def lookup_for_key_in_cache_or_database(settings, value,  key, default):
    return new_value

settings.dynaconf.add_hook(Action.AFTER_GET, lookup_for_key_in_cache_or_database)
```

Then in any part of code

```python
>>> from config import settings
>>> settings.MY_AWESOME_KEY
"doing things to lookup key 'MY_AWESOME_KEY' on cache or db"
```


rochacbruno:
replaced by #975

PR #842, opened 2022-12-18
Include tests in the sdist

gotmax23:


gotmax23:
There's a typo here. Let me fix it before reviewing.

Issue #853, opened 2023-01-10
[bug] Dynaconf fails with FileNotFoundError when current working directory doesn't exists


Wenzel:
**Describe the bug**
When dynaconf is invoked from a non existent current working directory (created and removed afterwards), the `dotenv` loading mechnism will fail with `FileNotFoundError`, even if `load_dotenv` has been disabled.

**To Reproduce**

~~~shell
python3 -m venv venv
source venv/bin/activate
pip install dynaconf
mkdir nonexistent_dir
cd nonexistent_dir
rm -r ../nonexistent_dir
dynaconf
~~~

**Expected behavior**
Dynaconf command line should have executed without any issues.

**Actual behavior**

`os.getcwd()` fails with FileNotFoundError
~~~python
Traceback (most recent call last):
  File "/home/mtarral/venv/bin/dynaconf", line 5, in <module>
    from dynaconf.cli import main
  File "/home/mtarral/venv/lib/python3.8/site-packages/dynaconf/__init__.py", line 3, in <module>
    from dynaconf.base import LazySettings  # noqa
  File "/home/mtarral/venv/lib/python3.8/site-packages/dynaconf/base.py", line 17, in <module>
    from dynaconf.loaders import env_loader
  File "/home/mtarral/venv/lib/python3.8/site-packages/dynaconf/loaders/env_loader.py", line 8, in <module>
    from dynaconf.vendor.dotenv import cli as dotenv_cli
  File "/home/mtarral/venv/lib/python3.8/site-packages/dynaconf/vendor/dotenv/cli.py", line 16, in <module>
    @click.option('-f','--file',default=os.path.join(os.getcwd(),'.env'),type=click.Path(exists=_A),help='Location of the .env file, defaults to .env file in current working directory.')
FileNotFoundError: [Errno 2] No such file or directory
~~~

**Proposed fix**
Dynaconf should suppress this `FileNotFoundError` for `os.getcwd()`

**Environment (please complete the following information):**
 - OS: Ubuntu 20.04
 - Dynaconf Version 3.1.11

Thanks !

PR #857, opened 2023-01-31
Handle all failures when pwd does not exist.

jctanner:
Fixes https://github.com/dynaconf/dynaconf/issues/853

Reproducer:

```bash
#!/bin/bash

rm -rf /tmp/foobar
mkdir -p /tmp/foobar
cd /tmp/foobar
rm -rf /tmp/foobar


echo "PWD:$(pwd)"

dynaconf list
```

### DIFF
--- a/dynaconf/loaders/env_loader.py
+++ b/dynaconf/loaders/env_loader.py
@@ -6,10 +6,11 @@ from dynaconf.utils import missing
 from dynaconf.utils import upperfy
 from dynaconf.utils.parse_conf import parse_conf_data
 
-DOTENV_IMPORTED=False
+DOTENV_IMPORTED = False
 try:
     from dynaconf.vendor.dotenv import cli as dotenv_cli
-    DOTENV_IMPORTED=True
+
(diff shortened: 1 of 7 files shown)


dynaconf/loaders/env_loader.py
```diff
@@ -5,7 +5,15 @@
 from dynaconf.utils import missing
 from dynaconf.utils import upperfy
 from dynaconf.utils.parse_conf import parse_conf_data
-from dynaconf.vendor.dotenv import cli as dotenv_cli
+
+DOTENV_IMPORTED=False
+try:
+    from dynaconf.vendor.dotenv import cli as dotenv_cli
+    DOTENV_IMPORTED=True
```
rochacbruno, line 12: where is this constant `DOTENV_IMPORTED` being referenced? shouldn't be on the `write` function at this same file?

jctanner: Forgot to add the check down in the `write` function, but it's there now.

rochacbruno (approved):

Issue #874, opened 2023-03-03
[bug] dynaconf.loaders.write does not support merge argument


Holzhauer:
**Describe the bug**
Despite its [documentation](https://www.dynaconf.com/advanced/#exporting) [loaders.write()](url) does not support the `merge` argument, which is available at `loader.write()`, where `loaders.write()` delegates to (it sets the argument to `False`),

**To Reproduce**
Having `settings.toml` in the root path.

Having the following app code:
 
```python
from dynaconf import Dynaconf
from dynaconf import loaders
from dynaconf import settings
from dynaconf.utils.boxing import DynaBox

settings = Dynaconf(settings_files=["settings.toml",])

data = settings.as_dict(env='development')
loaders.write('export.yaml', DynaBox(data).to_dict(), merge=False, env='development')
```

**Expected behavior**
Above code runs without error.

**Environment (please complete the following information):**
 - OS: Windows/10 Enterprise
 - Dynaconf Version: 3.1.11

rochacbruno:
similiar to #839 we need to fix it

pedro-psb:
This is working. Maybe we should add some tests?
If so, it should cover all core loaders, because each one of them should accept and handle `merge` correctly.

Refactor suggestion: It may be better to have the context (`loaders.write`) handle the merge option, so individual `loader.write` would just receive data and assume it was handled correctly. This would make it easier to test and harder to introduce bugs.

rochacbruno:
@pedro-psb the fix was introduced in https://github.com/dynaconf/dynaconf/pull/875/files but not released yet

You are suggesting to pass `**kwargs` to that function?

pedro-psb:
Not really. Indeed, I forget about a detail when I did the previous comments. Because merging requires the file to be opened first, this would require the outer context to have the specific file format readers/parsers. So maybe all the logic (read/parse + write to file) could be uplifted. At the end, it would be like a functional template method pattern for the module. I'll provide a sample of how this could work below.

Another way I thought (and prefer) was to turn loaders into classes (and then use regular template pattern). I think it would provide better interface for custom loaders too. By providing basically the `parser` and the `writter`, the custom loader would leverage dynaconf merging options for free. This would be a more long run proposal, maybe 4.0.0

The first proposal (simpler) would be like this:

 **Current (simplified)**
```python
# loaders/__init__.py
def write(filename, data, env, merge=False):
	loader = globals().get(loader_name)
	loader.write(filename, data, env, merge)

[...]

PR #875, opened 2023-03-03
Add merge argumento to `loader.write` to fix #839

### LINKED ISSUE #839: [bug][Documentation] Exporting: write() got an unexpected keyword argument 'merge'
**Describe the bug**
Following the example on the documentation to [export](https://www.dynaconf.com/advanced/#exporting) Dynaconf data to a file raises an exception with the `merge` argument

**To Reproduce**
~~~Python
loaders.write("/a/b/c", DynaBox(config).to_dict(), merge=False)
~~~

**Expected behavior**
The file should have been written

**Actual Behavior**
`TypeError: write() got an unexpected keyword argument 'merge'`

Just a quick documentation fix,
thanks !

Mdslino:


Mdslino:
closes #874

closes #839

dynaconf/loaders/__init__.py
```diff
@@ -263,7 +263,7 @@ def enable_external_loaders(obj):
             obj.LOADERS_FOR_DYNACONF.insert(0, loader)
 
 
-def write(filename, data, env=None):
+def write(filename, data, merge=False, env=None):
```
rochacbruno, line 266: I would add the `merge=False` as the last argument, because if someone is already relying on positional argument order it will break.

```suggestion
def write(filename, data, env=None, merge=False):
```

rochacbruno (changes requested):

dynaconf/loaders/__init__.py
Mdslino: I agree. I did that initially but changed it to reflect the exemplified use in the docs. I'll change.

PR #886, opened 2023-03-10
test 885

rochacbruno:
test #885

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,4 +1,4 @@
-# Dynaconf Quick Start
+# Dynaconf Quick Start!
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


rochacbruno:
@pedro-psb it works, but as you said now the required checkpoint will never be reached, not sure if there is a way to fix this or if we just remove the checkpoint and trust the reviewers/mergers

pedro-psb:
> @pedro-psb it works, but as you said now the required checkpoint will never be reached, not sure if there is a way to fix this or if we just remove the checkpoint and trust the reviewers/mergers

The docs provide a solution for that [here](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks). I'll make another pull request.

rochacbruno:
It works

PR #889, opened 2023-03-13
Add FAQ questions and small typos on Validation

pedro-psb:
About typos, I was reading the Validators docs and found one typo. Then I decided to use Grammarly to check for other things (which I wouldn't catch ever).


rochacbruno:
That is really nice, Thanks

PR #892, opened 2023-03-15
Fix / CLI command validate not working properly

### LINKED ISSUE #879: [bug] `dynaconf validate ` not working with `dynaconf_validators.toml`
### Discussed in https://github.com/dynaconf/dynaconf/discussions/868

<div type='discussions-op-text'>

<sup>Originally posted by **Holzhauer** February 14, 2023</sup>

> I attempt to validate the type of a variable in `dynaconf_validators.toml`. However, if I code it as
> 
> `startyear = {is_type_of=int}`
> 
> I get a `dynaconf.vendor.tomllib.TOMLDecodeError: Invalid value` for the character after "=". If I set the type in quotes like
> 
> `startyear = {is_type_of="int"}`
> 
> I get a `TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union`.
> What is the correct syntax here?

</div>

---

I'll describe the bug from the initial discussion but I think there is more room to improvement in the implementation of the `dynaconf validate`

### Describe the bug

When trying to cast a value to `int` using the cli tool `dynaconf validate` and the file `dynaconf_validators.toml`, there program raises a `TypeError`. Also, a wrong use of TOML format using unquoted `int` raises an unhandled `tomllib` specific error.

### To Reproduce

1) Download [gist sample](https://gist.github.com/pedro-psb/d26a78d2d3742b1086d935ba8e9bd109), setup a virtualenv, pip install `dyanconf`  and execute `dynaconf -i config.settings validate`. It should raise a TypeError.

Shortcut:

```bash
# get code
git clone https://gist.github.com/d26a78d2d3742b1086d935ba8e9bd109.git issue-879 && cd issue-879

# review and execute
cat run.sh
chmod +x run.sh
./run.sh
```

2) Using unquoted `int` raises some unhandled `tomllib` errors.

### Expected behavior

1) Validate the settings without any problem
2) Handle error to suggest it should probably be quoted.

### Environment

- OS: Ubuntu 22.04
- Python Version: 3.11.2
- Dynaconf Version: 3.1.12

---
### Additional comments

I've found that the cli command `validate` has some other bugs and could be better tested. Eg, if in this code sample the `environments` setting was false, and there were no `[default]` in the `.toml`, the output would be:

```bash
Invalid rule for parameter 'is_type_of'
Validation success!
```

I'd like to propose to write some more unit test for this command and try to fix these inconsistencies. Should I create another issue for that?

[edit: add gist sample code]

pedro-psb:
This PR does these:

- Fix #879 
  - Implements the preferred solution presented [here](https://github.com/dynaconf/dynaconf/issues/879#issuecomment-1462066996), which is using `"@type int"`.
  - adds support for using `"int"`, with a trivial modification of `type_conditions` (which can be easily reverted if this is not desired). I would make this a different commit, but I accidentally amended them and didn't know how to split back.
  - Replace raw tomllib parsing with existing loaders
  - Fix bug when iterating over a file which don't have `[default]` defined explicitly.
- Fix mypy complain on truthy conditional expression.

### DIFF
--- a/dynaconf/cli.py
+++ b/dynaconf/cli.py
@@ -724,19 +724,31 @@ def validate(path):  # pragma: no cover
         click.echo(click.style(f"{path} not found", fg="white", bg="red"))
         sys.exit(1)
 
-    # I added this Dynaconf instance to use the standard loaders/converters
-    # data processing require passing a settings instance.
-    # This can be avoided with some extra work
-    temp_settings = Dynaconf()
-    toml_loader.load(temp_settings, filename=str(path))
-    validation_data = temp_settings.as_dict()
-    temp_settings = None  # does this induce garbage collector? don't know
+    # parse validator file
+    try:  # try tomlib first
+        validation_data = tomllib.load(open(str(path), "rb"))
+    except UnicodeDecodeError:  # fallback to legacy toml (TBR in 4.0.0)
+        warnings.warn(
+            "TOML files should have only UTF-8 encoded characters. "
+            "starting on 4.0.0 dynaconf will stop allowing invalid chars.",
+        )
+        validation_data = toml.load(
+            open(str(path), encoding=default_settings.ENCODING_FOR_DYNACONF),
+        )
+    except tomllib.TOMLDecodeError as e:
+        click.echo(
+            click.style(
+                f"Error parsing TOML: {e}. Maybe it should be quoted.",
+                fg="white",
+                bg="red",
+            )
+        )
+        sys.exit(1)
 
     # guarantee there is an environment
-    if not validation_data.get(
-        "DEFAULT"
-    ):  # it is mandatory to have DEFAULT, right?
-        validation_data = {"DEFAULT": validation_data}
+    validation_data = {k.lower(): v for k, v in validation_data.items()}
+    if not validation_data.get("default"):
+        validation_data = {"default": validation_data}
 
     success = True
     for env, name_data in validation_data.items():
--- a/dynaconf/cli.py
+++ b/dynaconf/cli.py
@@ -12,9 +12,11 @@ from pathlib import Path
 
 from dynaconf import constants
 from dynaconf import default_settings
+from dynaconf import Dynaconf
 from dynaconf import LazySettings
 from dynaconf import loaders
 from dynaconf import settings as legacy_settings
+from dynaconf.loaders import toml_loader
 from dynaconf.loaders.py_loader import get_module
 from dynaconf.utils import upperfy
 from dynaconf.utils.files import read_file
@@ -713,7 +715,6 @@ def validate(path):  # pragma: no cover
     # reads the 'dynaconf_validators.toml' from path
     # for each section register the validator for specific env
     # call validate
-
     path = Path(path)
 
     if not str(path).endswith(".toml"):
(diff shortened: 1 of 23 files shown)


dynaconf/utils/parse_conf.py
```diff
@@ -228,6 +228,16 @@ def evaluate(settings, *args, **kwargs):
...
 
+# is there another way to make this conversion?
+type_map: dict[str, type] = {
+    "int": int,
+    "float": float,
+    "str": str,
+    "bool": bool,
+    "list": list,
+    "dict": dict,
+}
```
rochacbruno, line 239: this data is on `__builtins__`

```
In [2]: getattr(__builtins__, "int")
Out[2]: int

In [3]: getattr(__builtins__, "str")
Out[3]: str
```

dynaconf/utils/parse_conf.py
```diff
@@ -248,6 +258,11 @@ def evaluate(settings, *args, **kwargs):
     )
     if isinstance(value, Lazy)
     else json.loads(value),
+    "@type": lambda value: value.set_casting(
+        type
+    )  # don't know what this really does
+    if isinstance(value, Lazy)
+    else type_map.get(value),
```
rochacbruno, line 265: do we really need `@type` here?

With the following syntax working
```
should_be_number={is_type_of="int"}
```
we probably don't need to also support `@type`, we can assume that `is_type_of` accepts any type present on `__builtins__`

dynaconf/utils/parse_conf.py
```diff
@@ -248,6 +258,11 @@ def evaluate(settings, *args, **kwargs):
     )
     if isinstance(value, Lazy)
     else json.loads(value),
+    "@type": lambda value: value.set_casting(
+        type
+    )  # don't know what this really does
```
rochacbruno, line 263: To answer the question, this `if` here is used because it is possible to combine multiple converters like `@int @format {env[NUMBER]}` then it will first get the result from `@format ...` which will be a `str` wrapped in a `Lazy` and then we pass it again to the `int` converter.

dynaconf/validator_conditions.py
```diff
@@ -42,6 +42,13 @@ def identity(value, other):
 
 def is_type_of(value, other):
     """Type check"""
+
+    # auto transform quoted types
+    if isinstance(other, str):
+        from dynaconf.utils.parse_conf import type_map
+
+        other = type_map[other]
```
rochacbruno, line 50:
```suggestion
        other = getattr(__builtins__, other)
```
this should work for all builtin types like str, int, dict, set etc

dynaconf/cli.py
```diff
@@ -723,24 +724,28 @@ def validate(path):  # pragma: no cover
...
+    temp_settings = Dynaconf()
+    toml_loader.load(temp_settings, filename=str(path))
+    validation_data = temp_settings.as_dict()
+    temp_settings = None  # does this induce garbage collector? don't know
+
+    # guarantee there is an environment
+    if not validation_data.get(
+        "DEFAULT"
+    ):  # it is mandatory to have DEFAULT, right?
+        validation_data = {"DEFAULT": validation_data}
```
rochacbruno, line 739: what are we trying to solve by implementing this?

dynaconf/cli.py
```diff
@@ -723,24 +724,28 @@ def validate(path):  # pragma: no cover
...
+    # This can be avoided with some extra work
+    temp_settings = Dynaconf()
+    toml_loader.load(temp_settings, filename=str(path))
+    validation_data = temp_settings.as_dict()
+    temp_settings = None  # does this induce garbage collector? don't know
+
+    # guarantee there is an environment
+    if not validation_data.get(
+        "DEFAULT"
+    ):  # it is mandatory to have DEFAULT, right?
```
rochacbruno, line 738: For the validation file we can assume there will be always a `[default]` even if dynaconf is not using environments we can set that the file will always have it, or we inject it there as you did

dynaconf/utils/parse_conf.py
pedro-psb: Yes, that's reasonable.
@type probably wouldn't be used very much

dynaconf/validator_conditions.py
pedro-psb: Will do that

dynaconf/cli.py
pedro-psb: I've used the toml_loader because it would convert the `@type` for me.
I guess I can revert that now.

dynaconf/validator_conditions.py
pedro-psb: For some reason, this raises an error in the app context (even though it works on the interpreter). I've assured `other` is a string with value of 'int', so this is really weird.

```
    other = getattr(__builtins__, other)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'dict' object has no attribute 'int'
```

But this works:

```python
other = __builtins__.get(other)
```

pedro-psb:
Sorry for --force pushing, I guess this makes the PR messy.
I need to learn proper git for real.

tests_functional/specific_settings_files/Makefile
pedro-psb, line 1: This was an accident

rochacbruno (approved):

Issue #712, opened 2022-01-27
Generating Dynaconf object from a python dict


pvij:
**Is your feature request related to a problem? Please describe.**
I usually use JSON configs in python projects and upload them to AWS parameter store. I really like using dynaconf as it helps me in validating the config and converting the config to an object. The problem I face is that when I get the parameter from the AWS parameter store, I get a stringified JSON (which has to be extracted from the response). Now, to use dynaconf, I have to write the stringified JSON to a file and then use dynaconf. It seems to me that writing to a file is an unnecessary additional step.

**Describe the solution you'd like**
I would like a Dynaconf object to be created from a python dict structure.

**Describe alternatives you've considered**
* I can simply stop using dynaconf and use the dict directly (and try validating the dict on my own).

**Additional context**
* I tried searching for any similar issue that might already exist but I didn't find any. Please let me know if there is a similar issue I might have missed.
* If there is any advice you have regarding the way I am getting the config, please feel free to share it. Also, if my request doesn't make sense, it would be great if you could point out why with reason.

Thank you for reading the above.

rochacbruno:
Did you mean something like?

```py
from dynaconf import Dynaconf
settings = Dynaconf()

data = {"name": "Bruno", "info__age": 30, "info__id": 123}
settings.update(data)

In [6]: settings.name
Out[6]: 'Bruno'

In [7]: settings.info
Out[7]: <Box: {'id': 123, 'age': 30}>

In [8]: settings.info.id
Out[8]: 123

In [9]: settings.info.age
Out[9]: 30



```

pvij:
Got it.
```
from dynaconf import Dynaconf, Validator
config = Dynaconf() 
config.update(data, validators=[Validator("name", must_exist=True)])
```
Thank you for getting back to me :-)

rochacbruno:
maybe

```py
config.validators.register(Validator("aws", "elasticsearch", must_exist=True))
config.update(data)
config.validators.validate()
```

validators are not trigerred upon update, maybe they should... we can change it.

pvij:
Will have to go deeper into the source code (and also see how other similar libraries work) to comment if it should be changed. But the above works for now. Thank you for your replies.

rochacbruno:
We can add `validate=True` to methods `update`, `set`, `load_file`.

All the methods that can be called on-demand, so it is optional.

pedro-psb:
> We can add validate=True to methods update, set, load_file.
> 
[...]

PR #900, opened 2023-03-23
Feature - Validate on call to update, set or load_file

pedro-psb:
This add support for choosing whether to validate or not data loaded by calling the `Settings` methods `update`, `set` and `load_file`. It also adds a global option called `VALIDATE_ON_UPDATE_FOR_DYNACONF` which sets the default behaviour for validating on updates. Related to #712

## Sample usage

```python
settings = Dynaconf() # default is validate_on_update=False, so it won't have compat issues
settings.validators.register(Validator("value_a", must_exist=True))
settings.update({"value_b": "foo"}) # will bypass validation
settings.update({"value_b": "foo"}, validate=False) # will raise ValidationError
```

```python
settings = Dynaconf(validate_on_update=True)
settings.validators.register(Validator("value_a", must_exist=True))
settings.update({"value_b": "foo"}, validate=False) # will bypass validation
settings.update({"value_b": "foo"}) # will raise ValidationError
```

## Notes

- `load_file` required adding `validate` to lots of places where `set` and `update` were used by loaders. Should there be tests to each one of the loaders?
- I shall update the docs on this PR later if the code is approved


rochacbruno (approved):

pedro-psb:
Will this be shipped in the next release (3.1.13)? Just so I can add the right note about when it was introduced.

pedro-psb:
I've noticed that the current implementation does not include support for using `validate_all()`. It will always use `validate()`.

What do you think of adding this option to the interface?

```python
settings.update({...}, validate=True) # runs `validate`
settings.update({...}, validate=False) # does not run
settings.update({...}, validate="all") # runs `validate_all()`  <<<
```

rochacbruno:
I like the idea of adding validate_all just ensure to change the conditions to `if validate is True` instead of just `if validate` because `all` would also evaluate to True

pedro-psb:
I've finished this, but I'm having some trouble pushing the up-to-date work. This last commit was not supposed to be there, it is some side-effect of me trying to push the actual work.

I'll try to figure this out without using `--force`.

dynaconf/default_settings.py
```diff
@@ -246,6 +246,8 @@ def reload(load_dotenv=None, *args, **kwargs):
     "APPLY_DEFAULT_ON_NONE_FOR_DYNACONF", None
 )
 
+VALIDATE_ON_UPDATE_FOR_DYNACONF = get("VALIDATE_ON_UPDATE", False)
```
rochacbruno, line 249: I think the variable must be with suffix here

get function could be smart to add missing suffix but I think it needs now.

```suggestion
VALIDATE_ON_UPDATE_FOR_DYNACONF = get("VALIDATE_ON_UPDATE_FOR_DYNACONF", False)
```

Issue #837, opened 2022-11-25
Inspect and Log: Visibility into the settings file loading process ?


Wenzel:
Hi,

I would like to have more visibility into how Dynaconf is loading the settings files, and overriding the variables at each step of the process.
This would allow the uses of my app to clearly understand what went wrong in their environment.

For example:
~~~
Loading from /etc/app/config.toml
SET port = 443
Loading from /home/dynaconf/.config/app/config.toml
Not found
Loading from /home/dynaconf/project/.env
OVERRIDE port = 8080
SET protocol = https
~~~

Would it be possible to have that clarity into the loading process ?

Thanks for Dynaconf ! ❤️

rochacbruno:
Right all the loading history is saved on `_loaded_by_loaders` attribute, however all the logging has been removed on version 3.0 and we plan to add logging back in the 4.0 version with a better implementation.

rochacbruno:
The `inspect` functionality will be a feature added to 4.0, a beta version of the implementation may be added to 3.2.x

pedro-psb:
Inspect functionality was released today on `3.2.0` as an experimental feature. It ships an [utility function](https://www.dynaconf.com/advanced/#inspect_settings) and the counterpart [cli command](https://www.dynaconf.com/cli/#dynaconf-inspect-tech-preview).

It does not provide that exact form of output, but it offers some raw data (kind like a state-history for each source) that may be used to generate other kind of reports. Hope it can be useful :)

PR #901, opened 2023-03-23
replace bare tries with suppress

rochacbruno:


pedro-psb (approved):
I guess when logging is reimplemented (as #837) these trivial catches can be used to log.

Issue #858, opened 2023-02-04
Suggestions on how to `reverse` in my Django settings.yaml


wgordon17:
I currently define settings like `LOGIN_URL` as `reverse_lazy("account_login")`. Is there a way to replicate this type of configuration into something I can use in my Dynaconf settings.yaml?

If possible, I'd prefer to keep all settings in my Dynaconf settings.yaml and keep my Django settings.py file as empty as possible.

rochacbruno:
What you can do right now is the following:

```python
# on the top import reverse_lazy
from django.urls import reverse_lazy


Then after the `import dynaconf`  - before `settings = ...` add

```pythin
dynaconf.utils.parse_conf["@reverse_lazy"] = reverse_lazy
```


Now you should be able to do

On yaml
```yaml
default:
    login_url: "@reverse_lazy account_login"
```

On env vars
```bash
export DJANGO_LOGIN_URL="@reverse_lazy account_login"
```

**NOTE** I have not tested! but if it works we can then add it by default to the context in future versions.

wgordon17:
This is _excellent_, thank you! 🎉 

So, based on the format in https://www.dynaconf.com/envvars/#adding-a-custom-casting-token and from your suggestion, I came up with

```python
# HERE STARTS DYNACONF EXTENSION LOAD (Keep at the very bottom of settings.py)
# Read more at https://www.dynaconf.com/django/
import dynaconf  # noqa
from dynaconf.utils import parse_conf  # noqa

parse_conf.converters["@reverse_lazy"] = (
    lambda value: value.set_casting(reverse_lazy)
    if isinstance(value, parse_conf.Lazy)
    else reverse_lazy(value)
)
settings = dynaconf.DjangoDynaconf(__name__)  # noqa
# HERE ENDS DYNACONF EXTENSION LOAD (No more code below this line)
```

And then I was able to do exactly what you suggestion in my `settings.yaml`

```yaml
default:
    LOGIN_URL: "@reverse_lazy account_login"
```

Perfect! 👌

wgordon17:
I have _no_ idea what was going on when I first validated this, but I have _yet_ to be able to reproduce getting this working. After _waayyy_ too much time debugging this, I discovered that https://github.com/dynaconf/dynaconf/blob/master/dynaconf/base.py#L894 is the offending line breaking this functionality

```python
if existing is not None and existing != value:
```

[...]

PR #904, opened 2023-03-26
Provides new add_converter function for #858

wgordon17:
Provides new add_converter function for cases of adding other converters, like Django's `reverse_lazy`. Wraps the converter up and enables converter chaining, e.g., with `@format`. Includes a functional test and update to docs.

### DIFF
--- a/dynaconf/utils/parse_conf.py
+++ b/dynaconf/utils/parse_conf.py
@@ -276,6 +276,20 @@ def get_converter(converter_key, value, box_settings):
     return converted_value
 
 
+def add_converter(converter_key, casting_callable):
+    global converters
+
+    def converter_callable(value, box_settings=None):
+        if isinstance(value, Lazy):
+            # value got here as a Lazy object, needs evaluation
+            return Lazy(value(box_settings), casting=casting_callable)
+        else:
+            # value is good to go
+            return Lazy(value, casting=casting_callable)
+
+    converters[f"@{converter_key}"] = converter_callable
+
+
 def parse_with_toml(data):
     """Uses TOML syntax to parse data"""
     try:  # try tomllib first
--- a/dynaconf/utils/parse_conf.py
+++ b/dynaconf/utils/parse_conf.py
@@ -276,18 +276,20 @@ def get_converter(converter_key, value, box_settings):
     return converted_value
 
 
-def add_converter(converter_key, casting_callable):
-    global converters
+def add_converter(converter_key, func):
+    """Adds a new converter to the converters dict"""
+    if not converter_key.startswith("@"):
+        converter_key = f"@{converter_key}"
 
-    def converter_callable(value, box_settings=None):
-        if isinstance(value, Lazy):
-            # value got here as a Lazy object, needs evaluation
-            return Lazy(value(box_settings), casting=casting_callable)
-        else:
-            # value is good to go
-            return Lazy(value, casting=casting_callable)
-
-    converters[f"@{converter_key}"] = converter_callable
+    converters[converter_key] = wraps(func)(
+        lambda value: value.set_casting(func)
+        if isinstance(value, Lazy)
+        else Lazy(
+            value,
+            casting=func,
+            formatter=BaseFormatter(lambda x, **_: x, converter_key),
+        )
+    )
 
 
 def parse_with_toml(data):
--- a/tests_functional/django_example/settings.yaml
+++ b/tests_functional/django_example/settings.yaml
@@ -8,6 +8,7 @@ default:
   SECRET_KEY: 1234
   STATIC_ROOT: .
   STATIC_URL: /static/
+  LOGIN_URL: "@reverse_lazy admin:login"
   ALLOWED_HOSTS:
     - '*'
   INSTALLED_APPS:
--- a/tests_functional/django_example/settings.yaml
+++ b/tests_functional/django_example/settings.yaml
@@ -8,7 +8,8 @@ default:
   SECRET_KEY: 1234
   STATIC_ROOT: .
   STATIC_URL: /static/
-  LOGIN_URL: "@reverse_lazy admin:login"
+  ADMIN_NAMESPACE: admin
+  LOGIN_URL: "@reverse_lazy @format {this.ADMIN_NAMESPACE}:login"
   ALLOWED_HOSTS:
     - '*'
   INSTALLED_APPS:
--- a/docs/django.md
+++ b/docs/django.md
@@ -171,11 +171,10 @@ Example: if you need to use `reverse_lazy`, you might do this:
 # HERE STARTS DYNACONF EXTENSION LOAD (Keep at the very bottom of settings.py)
 
 import dynaconf
-from dynaconf.utils import parse_conf
 from django.urls import reverse_lazy
 
 # register custom converter
-parse_conf.converters["@reverse_lazy"] = lambda value: parse_conf.Lazy(value, casting=reverse_lazy)
+dynaconf.add_converter("reverse_lazy", reverse_lazy)
 
 # regular setup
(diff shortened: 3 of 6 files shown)


tests_functional/django_example/settings.yaml
pedro-psb, line 1: It would be good to test if it is composable by chaining it with some other converter.

pedro-psb (commented):
It would be good to test if it is composable too by chaining it with some other converter.

wgordon17:
@pedro-psb It _definitely doesn't seem composable. I made the following change in `settings.yaml

```yaml
  ADMIN_NAMESPACE: admin
  LOGIN_URL: "@reverse_lazy @format {this.ADMIN_NAMESPACE}:login"
```

And it blew up

```
TypeError: descriptor 'format' for 'str' objects doesn't apply to a 'Lazy' object
```

If I try to change the order and use `@format @reverse_lazy`, then the `@reverse_lazy` value just gets treated as raw text.

I'm open to any suggestions, otherwise we just call this a "noncomposable" parser? ¯\\\_(ツ)_/¯

pedro-psb:
@wgordon17 I'll try something out tomorrow, I don't know much about the converters' mechanism yet.

But yes, although it would be better to provide a composable example, it is already very useful to have a non-composable one.

wgordon17:
Sounds good @pedro-psb, let me know if I can help. Ideally, I wouldn't have to use `Lazy` at all, but I can't seem to figure out a way past early-evalution. Since reversing Django routes relies on a fully configured project, but you can't configure the project without settings being fully setup, and the settings can't get setup because the route is trying to be reversed...and around and around we go 😉 

For me, https://github.com/dynaconf/dynaconf/blob/master/dynaconf/base.py#L894, was the offending line that was causing early value evaluation breaking the "laziness" of `reverse_lazy`. But wrapping it in a `Lazy` object seemed to fix the early evaluation issue. Perhaps there's a _different_ way to wrap the parser/casting to still "act" like a `str` while still delaying evaluation ¯\\\_(ツ)_/¯

pedro-psb:
@wgordon17 I've found out that the caller of the converters (`get_converters`) optionally passes the `box_settings` to the converter functions.
Then, you can use it to evaluate the `value` if it got there as a Lazy object (because `@format` needs to be Lazy too).

```python
def reverse_lazy_fun(value, box_settings=None):
    if isinstance(value, Lazy):
        # value got here as a Lazy object, needs evaluation
        return Lazy(value(box_settings), casting=reverse_lazy)
    else:
        # value is good to go
        return Lazy(value, casting=reverse_lazy)


parse_conf.converters["@reverse_lazy"] = reverse_lazy_fun
```

rochacbruno:
@pedro-psb I think we must add a builtin helper to register new converters then we abstract this logic

```py
def reverse_lazy_fun(value, box_settings=None):
    if isinstance(value, Lazy):
        # value got here as a Lazy object, needs evaluation
        return Lazy(value(box_settings), casting=reverse_lazy)
    else:
        # value is good to go
        return Lazy(value, casting=reverse_lazy)
```

In something better to use like

```py
settings.dynaconf.add_converter("reverse_lazy", callable, combo=True)
```

pedro-psb:
@rochacbruno Yes, I think this would be good

rochacbruno:
This is what I was playing with @pedro-psb 
I will add to a PR later today, need to test with more scenarios.

Also I am not sure if keep it as a standalone function or bind it to the settings as a method

test_formater_helper

wgordon17:
@rochacbruno I was definitely not being elegant, but I just threw this together as well https://github.com/dynaconf/dynaconf/pull/904/files#diff-81216dc760c0d7472d00bab2b35a37d32985b9b0e07ffa9877ed2bbb664373c1

rochacbruno:
@wgordon17 do we really need to pass `box_settings` ? maybe your implementation is better

wgordon17:
@rochacbruno Heh, I really don't know 🙈 I only stole it from @pedro-psb's suggestion. I think you're the SME on the backend internals

dynaconf/utils/parse_conf.py
```diff
@@ -276,6 +276,20 @@ def get_converter(converter_key, value, box_settings):
     return converted_value
 
 
+def add_converter(converter_key, casting_callable):
+    global converters
```
rochacbruno, line 280: No need to be global here because `converters` is mutable

dynaconf/utils/parse_conf.py
```diff
@@ -276,6 +276,20 @@ def get_converter(converter_key, value, box_settings):
...
+
+    def converter_callable(value, box_settings=None):
+        if isinstance(value, Lazy):
+            # value got here as a Lazy object, needs evaluation
+            return Lazy(value(box_settings), casting=casting_callable)
+        else:
+            # value is good to go
+            return Lazy(value, casting=casting_callable)
```
rochacbruno, line 288: Would be good to set a `formatter` here, the default is `str.format` which works fine but we can avoid that with

```py
def noop_formatter_factory(converter_key):
    """Builds a new noop formatter and sets its token"""
    return BaseFormatter(lambda x, **_: x, converter_key)
```

and

```suggestion
            return Lazy(
                value, 
                casting=casting_callable, 
                formatter=noop_formatter_factory(f"@{converter_key}")
            )
```

dynaconf/utils/parse_conf.py
```diff
@@ -276,6 +276,20 @@ def get_converter(converter_key, value, box_settings):
     return converted_value
 
 
+def add_converter(converter_key, casting_callable):
+    global converters
+
+    def converter_callable(value, box_settings=None):
```
rochacbruno, line 282: We can use functools.wraps to make this generated function to inherit name and docstring from the passed function. https://docs.python.org/3/library/functools.html#functools.wraps

```suggestion
    @functools.wraps(casting_callable)
    def converter_callable(value, box_settings=None):
```

pedro-psb:
I think `add_converter` should be a module function, as it is modifying a module variable.

It could live in this same file (`parse_conf.py`), but as it is a internals module, it would be better to have a public `utils` or similar namespace just for users, so their autocompletion don't get crowded with plumbing stuff.

dynaconf/utils/parse_conf.py
```diff
@@ -276,6 +276,20 @@ def get_converter(converter_key, value, box_settings):
...
+        if isinstance(value, Lazy):
+            # value got here as a Lazy object, needs evaluation
+            return Lazy(value(box_settings), casting=casting_callable)
+        else:
+            # value is good to go
+            return Lazy(value, casting=casting_callable)
+
+    converters[f"@{converter_key}"] = converter_callable
```
rochacbruno, line 290: Just a suggestion, it is in practice the same, so any of the options might work

```suggestion
def add_converter(converter_key, func):
    """Adds a new converter to the converters dict"""
    if not converter_key.startswith("@"):
        converter_key = f"@{converter_key}"
    converters[converter_key] = wraps(func)(
        lambda value: value.set_casting(func) if isinstance(value, Lazy)
        else Lazy(
            value,
            casting=func,
            formatter=BaseFormatter(lambda x, **_: x, converter_key)
        )
    )

```

rochacbruno:
then on `dynaconf/__init__.py` we can do

```py
from dynaconf.utils.parse_conf import add_converter

__all__ = [
    ...
   "add_converter"
]
```

then the usage would be

```py
from dynaconf import Dynaconf, add_converter

settings = Dynaconf()
add_converter("key", function)
```

wgordon17:
Rebased and refactored with suggestions, thanks so much @pedro-psb @rochacbruno! Also included a quick change to the docs. Once you're good with this, I can squash all the commits down (or GitHub can do it ¯\\\_(ツ)_/¯)

rochacbruno (approved):

PR #910, opened 2023-04-03
Fix - Array entries duplicated when using Validator with default for computed value

### LINKED ISSUE #905: [bug] array entries duplicated with Validator using default for computed value.
**Describe the bug**
An array in the env file get strangely duplicated when using a `Validator` to create a new entry with a computed value.

It seems related to the presence of a string or an integer as array members, and to the `merge_enabled` flag.

**To Reproduce**

1. Having the following folder structure

<details>
<summary> Project structure </summary>

```bash

    .env
    .gitignore
    config.py
    settings.toml

```
</details>

2. Having the following config files:

<details>
<summary> Config files </summary>

**.env**
```bash
DYNACONF_GROUP__TEST_LIST = ["'1'", "'2'"]   # double quotes to cast as string

DYNACONF_GROUP__ANOTHER = 45
```

and

**settings.toml**
```toml
[default.group]
another_setting = "ciao"
```

</details>

3. Having the following app code:

<details>
<summary> Code </summary>

**config.py**
```python
from dynaconf import Dynaconf, Validator

settings = Dynaconf(
    envvar_prefix="DYNACONF",
    settings_files=["settings.toml"],
    environments=True,
    load_dotenv=True,
    merge_enabled=True,
)


settings.validators.register(
    Validator("group.something_new", default=5),
)

```

</details>

4. Executing under the following environment

<details>
<summary> Execution </summary>

```bash
dynaconf -i config.settings list
```

</details>

**Expected behavior**
This output:
```bash
Working in development environment
LOAD_DOTENV<bool> True
GROUP<dict> {'ANOTHER': 45,
 'TEST_LIST': ['1', '2'],
 'another_setting': 'ciao',
 'something_new': 5}
```

**What happens instead**
This output:
```bash
Working in development environment
LOAD_DOTENV<bool> True
GROUP<dict> {'ANOTHER': 45,
 'TEST_LIST': ["'1'", "'2'", '1', '2'],
 'another_setting': 'ciao',
 'something_new': 5}
```

 - OS: Windows 10x64
 - Dynaconf Version 3.1.12

**Additional context**
If you remove the `merge_enabled` flag, the problem does not occur. I know the flag is meaningless here but in more complex cases where it is needed the problem occurs identically.

I might be doing something wrong with the casting to string of those array elements, but I have tried some different ways and it still doesn't work (like with the `@str` format specifier).

pedro-psb:
This fixes #905

---

Just for the records, I'll try to explain what I've understood about the problem, sorry if it is too confusing.

One of the key-point of the problem was that the process of setting the default for the Validator uses `tomlfy=True` and this causes some side-effect on the `default` field siblings.

In the last step of recording this value there is a `merge_before_set(new, old)` function. I would expect that the new data to be merged in the existing data would be only the `default` value provided in Validation, but all of it's siblings are used too (including list).

Before this merge, `parse_conf_data(data, tomlfy=True, ...)` is called with 'default' value and all of it's existing siblings, and this makes `new` and `old` data to have different version of `test_list`, one that is "tomlyfied" and the other that is not (the original one).

### DIFF
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -864,6 +864,9 @@ class Settings:
                 key, value, loader_identifier=loader_identifier, tomlfy=tomlfy
             )
 
+        # Fix for #905
+        saved_value = value if loader_identifier == "setdefault" else None
+
         value = parse_conf_data(value, tomlfy=tomlfy, box_settings=self)
         key = upperfy(key.strip())
         existing = getattr(self, key, None)
@@ -896,9 +899,13 @@ class Settings:
             if merge:
                 value = object_merge(existing, value)
             else:
-                # `dynaconf_merge` may be used within the key structure
-                # Or merge_enabled is set to True
-                value = self._merge_before_set(existing, value)
+                # Fix for #905
+                if loader_identifier == "setdefault":
+                    value = saved_value
+                else:
+                    # `dynaconf_merge` may be used within the key structure
+                    # Or merge_enabled is set to True
+                    value = self._merge_before_set(existing, value)
 
         if isinstance(value, dict):
             value = DynaBox(value, box_settings=self)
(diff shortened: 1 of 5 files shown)


aqc-carlodri:
Thanks for the fix @pedro-psb!

PR #918, opened 2023-04-12
remove codecov

rochacbruno:


pedro-psb (approved):
I would forget to remove from all those places hehe

Issue #919, opened 2023-04-18
[bug] AttributeError is thrown if one of the configuration keys is an integer


ogajduse:
**Describe the bug**
Dynaconf fails to validate settings if any of the configuration keys are integer. Wrapping the key in quotes fixes the issue, but I would prefer dynaconf to keep its behavior from `3.1.11` where it did handle the integer-like keys correctly without throwing an exception.

**To Reproduce**
* have dynaconf 3.1.12 installed, I could not reproduce the bug with 3.1.11.
* initialize LazySettings
* validate settings

Project: [robottelo](https://github.com/SatelliteQE/robottelo/blob/b4b471aac102cfaeb40cffda7cd8452f1a2c1e1f/robottelo/config/__init__.py#L34)

Having the following config files:

<details>
<summary> Config files </summary>

**conf/ldap.yaml**
```yaml
LDAP:
  HOSTNAME:
    2016: env-ad-2016.acme.com
    2019: env-ad-2019.acme.com
  USERNAME: foobar
  PASSWORD: '@format {this.vault_awesome_password}'
  BASEDN: cn=Users,dc=department1,dc=loc1,dc=acme,dc=com
  GRPBASEDN: dc=department1,dc=loc1,dc=acme,dc=com
  REALM: INFRA.DEPARTMENT1.ACME.COM
  NAMESERVER:
    2016: 10.22.33.42
    2019: 10.22.33.43
  WORKGROUP:
    2016: SUBDEPT1
    2019: SUBDEPT12019
```

</details>

3. Having the following app code:
* [link to the repository](https://github.com/SatelliteQE/robottelo/blob/b4b471aac102cfaeb40cffda7cd8452f1a2c1e1f/robottelo/config/__init__.py#L19-L34)

4. Executing under the following environment

<details>
<summary> Execution </summary>

```bash
$ python -c 'from robottelo.config import settings'
```

</details>

**Expected behavior**
Dynaconf not throwing an AttributeError. Dynaconf should handle keys that are not strings as it did in version `3.1.11`.

**Environment (please complete the following information):**
 - OS: Fedora 36
 - Python 3.11
 - Dynaconf Version: 3.1.12
 - Frameworks in use: none

PR #920, opened 2023-04-18
Fix AttributeError with integer keys (yaml) #919

pedro-psb:
I'm not sure why this was supported before `v3.1.12` (see #919), because I didn't find explicit support for integer keys.

Anyway, the `ruamel.yaml` parser doesn't cast key integers to string (as toml parser does, for example), and some part of the system assumes all keys are strings, which I guess is a sane choice.

I've considered two approaches, assuming all int keys should be strings internally:
- Casting keys to integers in the loader level. This would require hacking the yaml parser (I guess this is not customizable) or writing a pos-parser to cast keys to string (this would cause some overhead, as all data would need to be traversed again).
- Cast keys when parsing them internally (`Settings.set`, `Settings.parse_conf`, etc). This wouldn't add overhead because the data need to pass through anyway. That's what I choose.

### Additional changes

I've also removed `pytest-dynaconf` recommendation on the main page, as their pypi link is broken.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -666,12 +666,6 @@ pip install dynaconf[redis]
 
 Read more on [external loaders](/advanced/#creating-new-loaders)
 
-## Extensions
-
-Community extensions will be linked here (open for recommendations).
-
-- [pytest-dynaconf]() - Set your Dynaconf environment to testing when running pytest.
-
 ## License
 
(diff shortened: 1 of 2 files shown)


rochacbruno:
I don't think we support integer keys befause those are not valid Python Identifiers and we need to keep dot notation working.

Issue #927, opened 2023-04-26
[RFC] Support for username and password with vault


hansharhoff:
**Is your feature request related to a problem? Please describe.**
Dynaconf currently does not support connecting to Hashicorp Vault using username and password.

**Describe the solution you'd like**
I suggest adding support for username and password.

**Describe alternatives you've considered**

**Additional context**
hvac already supports username and password and I've tested an implementation locally and could provide pull request if you like the approach.

pedro-psb:
Thanks for the contribution @hansharhoff !

PR #928, opened 2023-04-27
Vault userpass

hansharhoff:
As mentioned in #927 I propose to add userpass support for the vault loader.
Additionally, I have added two exception handling areas to handle the case where the list permission is not granted.


pedro-psb (commented):
This seems fine to me, but as I don't know much about Vault, I'll leave the approval to @rochacbruno

rochacbruno (approved):

pedro-psb:
The test is failure is unrelated to the PR.
It is failing because the newer version of `ipython` (8.13.0) only support `python>=3.9`. Pinning it to `8.12.1` in `requirements_dev.txt` should fix it. (`8.12.1` is the latest before the breaking one, and it will still receive some bugfixes.)

hansharhoff:
I have pinned the ipython dep.

pedro-psb (approved):

Issue #851, opened 2023-01-04
[bug] dynaconf is not installable in venvs without setuptools


jaklan:
### Bug description

`dynaconf` doesn't define any dependencies (due to vendoring), but that's not really true, because there is one runtime dependency - `pkg_resources` distributed with `setuptools`:
https://github.com/dynaconf/dynaconf/blob/0439bf836f1a22e96e4c71d388c2e68fd9b70425/dynaconf/contrib/flask_dynaconf.py#L17

How is it possible that it actually works? Only thanks to a "de facto" standard of pre-installing `setuptools` a) together with Python interpreter b) when [creating virtual environments](https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments):
>[venv](https://docs.python.org/3/library/venv.html) is available by default in Python 3.3 and later, and installs [pip](https://packaging.python.org/en/latest/key_projects/#pip) and [setuptools](https://packaging.python.org/en/latest/key_projects/#setuptools) into created virtual environments in Python 3.4 and later.
>
>[virtualenv](https://packaging.python.org/en/latest/key_projects/#virtualenv) needs to be installed separately, but supports Python 2.7+ and Python 3.3+, and [pip](https://packaging.python.org/en/latest/key_projects/#pip), [setuptools](https://packaging.python.org/en/latest/key_projects/#setuptools) and [wheel](https://packaging.python.org/en/latest/key_projects/#wheel) are always installed into created virtual environments by default (regardless of Python version).

It means `setuptools` was not explicitly declared, but it was assumed it would be there anyway. But as I mentioned - it never was a real standard and it caused serious issues in the Python build system as precisely described in [PEP 518](https://peps.python.org/pep-0518/#rationale):
>But when a project chooses to use setuptools, the use of an executable file like setup.py becomes an issue. You can’t execute a setup.py file without knowing its dependencies, but currently there is no standard way to know what those dependencies are in an automated fashion without executing the setup.py file where that information is stored. It’s a catch-22 of a file not being runnable without knowing its own contents which can’t be known programmatically unless you run the file.

And that's why `PEP 518` introduced the [build-system table](https://peps.python.org/pep-0518/#build-system-table) defined in `pyproject.toml`:
```
[build-system]
# Minimum requirements for the build system to execute.
requires = ["setuptools", "wheel"]  # PEP 508 specifications.
```
[...]

PR #934, opened 2023-05-10
Replace pkg_resources with importlib.metadata

thijsmie:
This PR removes the dependency on pkg_resources and relies on stdlib importlib.metadata instead. This fixes the problem described in issue #851.

### DIFF
--- a/dynaconf/contrib/flask_dynaconf.py
+++ b/dynaconf/contrib/flask_dynaconf.py
@@ -221,7 +221,9 @@ class DynaconfConfig(Config):
 
         for object_reference in app.config[key]:
             # parse the entry point specification
-            entry_point = EntryPoint(name=None, group=None, value=object_reference)
+            entry_point = EntryPoint(
+                name=None, group=None, value=object_reference
+            )
             # dynamically resolve the entry point
             initializer = entry_point.load()
             # Invoke extension initializer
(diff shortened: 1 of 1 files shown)


pedro-psb (commented):
Having a look at importlib docs, this seems fine.

Nevertheless, there are some other places where it is used (at vendor packages).
```
dynaconf/contrib/flask_dynaconf.py
17:import pkg_resources
226:            entry_point = pkg_resources.EntryPoint.parse(entry_point_spec)

dynaconf/vendor/click/decorators.py
282:                    import pkg_resources
286:                    for dist in pkg_resources.working_set:

dynaconf/vendor/ruamel/yaml/parser.py
74:# need to have full path with import, as pkg_resources tries to load parser.py in __init__.py

dynaconf/vendor/ruamel/yaml/setup.py
428:                        'import pkg_resources\n' 'pkg_resources.declare_namespace(__name__)\n'

```

thijsmie:
Hi @pedro-psb,

What is still needed to get this merged from my side? As far as I can tell the vendored import in click is protected by a try/except importerror construction so it has no real impact on usage. The ruamel yaml usage is code generation for namespace packages (depracated way of building namespace packages anyway). This is not a direct import either, no impact on usage of dynaconf as far as I can tell.

pedro-psb:
@thijsmie I see, thanks for the clarification.
I guess this is done then, but I'm not comfortable enough with packaging to merge it.
Let's wait @rochacbruno for final approval.

pedro-psb (approved):

rochacbruno:
@pedro-psb I will upgrade the vendored packages as looks like this is also fixed there

PR #937, opened 2023-05-24
Update validation.md to fix typo

gwynevans:
The doc has ValidationError being in dynaconf.validators not dynaconf.validator


gwynevans:
Ultra minor, but I came across it when prototyping without an IDE!

rochacbruno (approved):

PR #939, opened 2023-05-26
Feature - Inspect and CLI

pedro-psb:
Inspect functionality (related to #837).
The goal is to provide a way for the users to visualize the "loading-history" of any config key-path.

## Changes
- Replace `Settings._loaded_by_loaders` key type from `str` to `SourceMetadata<NamedTuple>` to store more data about each load, namely:
  - `loader`: the loader type (envvar_global, yaml, redis, vault..)
  - `identifier`: a loader identifier (eg, for file types it is the filename), for other it may be `unique`.
  - `env`: the env in which the data was loaded
  - `merged`: if the data was processed with some merge operation.
- Replace old `loader_identifier` format (`str`) with the new one (`SourceMetadata`) in most data-loading processes (calls to `set`).
- Implement inspecting utilities: `inspect_setting` (dumping) and `get_history` (direct access to python data).
- Implement new CLI command `inspect`, which basically ports `inspect_settings` to CLI.

## Tasks

- [x] **compatibility decision**: old `loaded_by_loaders` (`dict[str, dict]`) won't be kept, unless someone requests
- [x] preliminary tests/implementation of inspect:
- [x] refine inspect output:
  - [x] parametrize output stream (eg, stdout or regular file)
  - [x] transform `DynaBox` and `BoxList` into regular `dict|list` (for pretty-dumping) 
  - [x] parametrize output format/dumper (yaml, toml, python, etc)
- [x] Instance tests:
  - [x] get_history (cover mixing of envvar, file, validators and merges)
  - [x] inspect_settings
  - [x] integration tests (redis/vault).
- [x] CLI
  - [x] define it's usage/interface.
  - [x] implement and test (just basic usability, behaviour is covered in instance tests).
  - [x] fix test isolation and tmp_files setup (works running individually but not running the suite)
  - [x] handle errors properly
- [x] Write docs (`inspect_settings` and CLI)
- [x] Test coverage misses

Related ideas:
- star out secrets (`vault`,  `.secrets`, any other?)
- `inspect_settings` env filter will always include values from `default` env in active_value. It would be better to filter those out from `active_value` too.
- file ovewrite safe-check (disable overwrite by accident): for `inspect_settings` (`overwrite` flag param)
- ~~Report `.env` specific file identifier (it is loaded as `env_global`).~~

## Sample output

```yaml
header:
  filters:
    env: None
    key: foo
    history_ordering: ascending
  active_value: from_environ
history:
- loader: yaml
  identifier: /tmp/pytest-of-pedro-psb/pytest-84/test_inspect_print_key0/a.yaml
  env: default
  merged: false
  value:
    FOO: from_yaml
- loader: env_global
  identifier: unique
  env: global
  merged: false
  value:
    FOO: from_environ
    BAR: environ_only
```

## CLI usage

https://github.com/dynaconf/dynaconf/pull/939#issuecomment-1572708184

## Obs
- The new `SourceMetadata` should use more memory than the previous one, because it stores all data from all sources. It may be worth taking a look into that.
- `get_history`: envs that are no built-in names (default, development, global) won't be loaded if it is not the current env. Eg, a value from env `production` won't show in history unless `production` is the current env or is activated (with `settings.from_env` or similar). This is because of how env loading works.

## Testing

Current tests:

```bash
pytest tests/test_inspect.py -svk [testname]
pytest tests/test_cli.py -svk [testname]

### CODE REVIEW

**Comment by rochacbruno** [comment]
Dynaconf `inspect` CLI sub-command takes a key name (or dotted path) and outputs,
we have to keep it in conformity with existing CLI commands such as `list` so the options you listed are right.

```console
$ dynaconf inspect --help
Usage: dynaconf inspect [OPTIONS]

Inspect the current value and loading history for settings.

Options:
    -k, --key TEXT            Filters a single key
    -e, --env string          filter by environement. Eg: development
    -f, --format string       output format: json, yaml, table (default: table)
                              "table": prints the output to stdout     
    --help                    Show this message and exit.
```

```console
$ dynaconf inspect -k foo -f json
{
    "current": ...,
    "history": ...,
}
```
To save a file users can pipe the results or redirect stdout

```console
dynaconf inspect -k foo -f json | jq
dynaconf inspect -k foo -f json > file.json
```

> If `-k` is omitted then it dumps all defined keys, this would be useful to grep for filenames.
---

Programmatically the `inspect` function can take `key` as optional, if not passed it may dump all the keys.

> Also we must rename `inspect` to avoid conflict with python `inspect` module.

```py
import inspect
```

So it is better to rename

```python
from dynaconf.utils import inspect_settings

data = inspect_settings(settings)  # a full dict with all keys included
foo_data = inspect_settings(settings, "foo")  # a full dict with only "foo" history
```

pedro-psb:
@rochacbruno what does the "table" format means exactly? Could you provide a snippet of how the following should be rendered?

```yaml
header:
  filters:
    env: None
    key: foo
    history_ordering: ascending
  active_value: from_environ
history:
- loader: yaml
  identifier: /tmp/pytest-of-pedro-psb/pytest-84/test_inspect_print_key0/a.yaml
  env: default
  merged: false
  value:
    FOO: from_yaml
```

rochacbruno:
On `list` subcommand we are not doing any special treatment on secrets, that is understood that the person running this CLI has full access to the settings environment so we can output secrets or maybe if possible add an option like `--hide-secrets` to replace with ***** 

I am not sure it is easy to detect if a variable is a secret, maybe we can by detecting the source being .secrets. or vault.

pedro-psb:
I guess it would be good to star them out, as someone may want to share this report or something. Not very likely, but this should be easy to do by tracking `.secrets` or `vault` loader source.

pedro-psb:
> Does not work with a custom loader

@rochacbruno I've cheked the example at `tests_functional/custom_loader/` and the custom loader doesn't use BaseLoader, so it would have to copy how BaseLoader implements history recording (creating a `SourceMetadata` and passing to the obj  `set` or `update` method).
```diff
# tests_functional/custom_loader/my_custom_loader/sff_loader.py:load
+ source_metadata = SourceMetadata('sff', found_file, env)

if key:
    value = data.get(key.lower())  # sff format have lower case keys
+    obj.set(key, value, loader_identifier=source_metadata)
-    obj.set(key, value)
else:
+    obj.update(data, loader_identifier=source_metadata)
-    obj.update(data)
```

We could:
- Update the example in the [docs](https://www.dynaconf.com/advanced/?h=custom+loader#creating-new-loaders) and in the test to support history.
- Create a custom loader functional_test using BaseLoader (which should work, as builtins use it) and present it as the preferred way in the docs as well.

What do you think?

rochacbruno:
I think we need to update the docs and test to support history, there are projects out there using custom loaders based on the docs example, so we can point people on how to adjust their loaders to save history.

rochacbruno (approved):
🚀 lets merge it! awesome work!

Issue #690, opened 2021-11-15
[bug] Dynamic variables are evaluated before environment variables populate values


JacobCallahan:
**Describe the bug**
When using dynamic variables, we have noticed that they receive the value set in the settings file(s) and are not updated when those referenced settings are overridden by an environment variable.

**To Reproduce**

my_settings.yaml
```yaml
SERVER:
  VERSION:
    RELEASE: "6.10"
    SNAP: 22
  DEPLOY_WORKFLOW: "deploy-sat-jenkins"
  DEPLOY_ARGUMENTS:
    deploy_sat_version: "@format {this.server.version.release}"
    deploy_snap_version: "@format {this.server.version.snap}"
    deploy_rhel_version: '7'
```

my_app.py
```python
from dynaconf import Dynaconf

settings = Dynaconf(
    envvar_prefix="MYAPP",
    core_loaders=["YAML"],
    settings_file="my_settings.yaml",
)
```

loading my_app in ipython **without** environment variable
```
ipython -i my_app.py 
...
In [1]: settings.server.deploy_arguments
Out[1]: <Box: {'deploy_sat_version': '6.10', 'deploy_snap_version': '22', 'deploy_rhel_version': '7'}>

```

loading my_app in ipython **with(* environment variable. notice how the non-dynamic setting is correctly reflecting the value of the environment variable.
```
MYAPP_SERVER__VERSION__RELEASE="7.2" ipython -i my_app.py 
...
In [1]: settings.server.deploy_arguments
Out[1]: <Box: {'deploy_sat_version': '6.10', 'deploy_snap_version': '22', 'deploy_rhel_version': '7'}>

In [2]: settings.server.version.release
Out[2]: 7.2
```

**Expected behavior**
The dynamic variable should be evaluated **after** environment variables populate settings. This will allow for people to use both in conjunction with one another.

**Environment (please complete the following information):**
 - OS: Fedora 35
 - Python: 3.9.7
 - Dynaconf Version: 3.1.7

artctv:
Similar to my issue #603

rochacbruno:
Similar issues to be covered  by the fix of this issue:

https://github.com/rochacbruno/dynaconf/issues/603 
https://github.com/rochacbruno/dynaconf/issues/575
https://github.com/rochacbruno/dynaconf/issues/666

Issue #666, opened 2021-10-07
[bug] @format sometimes does not evaluate correctly when using environments


ZianVW:
**Describe the bug**
The @format dynamic variable does not produce the expected result with overlaid environments.  It seems that the unexpected behavior is triggered when there is another key on the same level of the settings tree.

**To Reproduce**

1. Having the following folder structure

The unexpected behavior occurs regardless of the folder structure.  I've attached a zip with source code necessary to reproduce the error, which can be unzipped into a single folder.

<details>
<summary> Project structure </summary>

```bash
.env
.gitignore
.secrets.yaml
config.py
main.py
requirements.in
requirements.txt
settings.yaml
```
</details>

2. Having the following config files:

<details>
<summary> Config files </summary>

**/path/.env**
```bash
ENV_FOR_DYNACONF=env1
```

and

**/path/settings.yaml**
```yaml
default:
    a_level1:
        prefix: default
        level2:
            value: "@format {this.a_level1.prefix}-xxx"
    b_level1:
        prefix: default
        level2:
            some_key: True
            value: "@format {this.b_level1.prefix}-xxx"

env1:
    dynaconf_merge: True
    a_level1:
        prefix: env1

    b_level1:
        prefix: env1
        level2:
            some_key: False
```

</details>

3. Having the following app code:

<details>
<summary> Code </summary>

**/path/src/main.py**
```python
from config import settings

print(f"a_level1.level2.value: {settings.a_level1.level2.value}")
print(f"b_level1.level2.value: {settings.b_level1.level2.value}  #<--- Evaluated to wrong value.  Expected: env1-xxx")
```

</details>

4. Executing under the following environment

<details>
<summary> Execution </summary>

Run using Python 3.8 in a virtual environment on Windows, with only dynaconf 3.1.7 installed.

$ python /path/src/main.py
```
PS C:\Users\dj36sv\Develop\scratch\dynaconf> python main.py
a_level1.level2.value: env1-xxx
b_level1.level2.value: default-xxx  <--- Evaluated to wrong value.  Expected: env1-xxx
```
</details>

**Expected behavior**
There are two cases in the provided settings file, under "a_level1" and "b_level1" respectively.

The case under "a_level1" works as expected.

[...]

Issue #603, opened 2021-06-24
[bug] Use of the format marker occurs before overriding values through an environment variable


artctv:
I'm not sure if this is a bug, but such logic is not described:
in `settings.toml` with setting `environments=True`:

```
[default.database]
DB_HOST = "127.0.0.1"
DB_PORT = "5432"
DB_ADDRES = "@format {this.database.DB_HOST}:{this.database.DB_PORT}"
```
Then, if the value `DB_PORT` will be overridden using the environment variable, `DB_PORT` value will be equal to the value from the environment variable, but the `DB_ADDRES` value will be created from the original `DB_PORT` equal `5432`

As I understand it, all variables using the marker `@format` occur before being overridden using environment variables. But, since the marker `@format` is a dynaconf feature and has nothing to do with the `.toml` format, this logic is not described anywhere, and can this be changed?

rochacbruno:
The lazy interpolation must occur later in the process when all the loaders have been loaded.

If not then this is a bug.

Issue #575, opened 2021-04-29
[bug] Variables in @format strings don't get overwritten


joshgordon:
**Describe the bug**
When using @format to define keys using other variables not in the root of the TOML file (or inside `[default]`), and overriding the variable being referenced in the @format string (e.g. by settings environment variables), the original variable gets used. 

**To Reproduce**
Have a config file like this: 

```
INSTALL_DIR = "/opt/myapp"
LOGS_DIR = "@format {this.INSTALL_DIR}/logs"

[ingest]
BASE_PATH = "/opt/ingest"
LOGS_DIR = "@format {this.ingest.BASE_PATH}/logs"
```

and run a python shell, and override the ingest.BASE_PATH with a different directory: 

$ DYNACONF_INGEST__BASE_PATH=/tmp/ingest python
>>> from config import settings
>>> settings.ingest.LOGS_DIR
'/opt/ingest/logs'
>>> settings.ingest.BASE_PATH
'/tmp/ingest'

Because seeing is better than explaining, [here's](https://replit.com/@joshgordon2/dynaconfplayground) a repl.it that demonstrates the issue. 

**Expected behavior**
The ingest.LOGS_DIR should inherit the new ingest.BASE_PATH brought in from the environment. Instead, it's using the base path as defined in the config file. 

**Environment (please complete the following information):**
 - OS: Centos 7
 - Dynaconf Version: 3.1.4
 - Frameworks in use (Flask, Django? versions..) N/A

**Additional context**
Git repo of the repl.it available [here](https://github.com/joshgordon/dynaconf_format_issue)

PR #944, opened 2023-06-29
Fix - Template substitution with variable update

pedro-psb:
This fixes the misbehavior of template substitution when the variable to be substituted gets updated by another source, as described in #575 #603 #666 and #690

**Main changes:**
- avoid replacement of `Lazy` objects by it's value inside `parse_conf_data`
- add `_safe_get` and `_safe_items` methods to `DynaBox`.
- small unrelated refactors

## Problem Overview

There were some points where earlier evaluation was called internally. Although those calls were unnecessary, they shouldn't have cause this bug by itself because evaluating a Lazy object means only asking it to return the final value (it doesn't replace itself with the value), so later on it should have used the correct value. Nevertheless, `parse_conf_data` was replacing the Lazy object with its early evaluation value, and that was the main cause of the problem.

## Additional Notes

Related to this, it would be nice to:
- solve previous related problems more cleanly (eg, #910 #933)
- track and remove unnecessary internal recursive evaluation calls (may improve performance).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -216,7 +216,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     services:
       vault:
-        image: vault
+        image: hashicorp/vault:latest
         ports:
           - 8200:8200
         env:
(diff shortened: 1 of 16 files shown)


pedro-psb:
CI failure seems related to `actions/checkout`: https://github.com/actions/checkout/issues/417
I'll try bumping v2 to v3 first. If this doesn't work, I'll explore the other solutions proposed there.

rochacbruno (approved):

Issue #648, opened 2021-08-26
[bug] Can't use gettext_lazy in django settings.py (provide an alternative)


rochacbruno:
```py
from django.utils.translation import gettext_lazy as _

# Default language
LANGUAGE_CODE = 'en-us'

LANGUAGES = [
    ('ja', _('Japanese')),
    ('nl', _('Dutch')),
    ('fr', _('French')),
    ('es', _('Spanish')),
    ('zh-hans', _('Chinese')),
    ('en-us', _('English (USA)')),
]
```

```bash
Traceback (most recent call last):
  File "/venv/lib64/python3.8/site-packages/django/utils/translation/trans_real.py", line 194, in _add_installed_apps_translations
    app_configs = reversed(list(apps.get_app_configs()))
  File "/venv/lib64/python3.8/site-packages/django/apps/registry.py", line 145, in get_app_configs
    self.check_apps_ready()
  File "/venv/lib64/python3.8/site-packages/django/apps/registry.py", line 136, in check_apps_ready
    raise AppRegistryNotReady("Apps aren't loaded yet.")
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/venv/bin/django-admin", line 8, in <module>
    sys.exit(execute_from_command_line())
  File "/venv/lib64/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/venv/lib64/python3.8/site-packages/django/core/management/__init__.py", line 363, in execute
    settings.INSTALLED_APPS
  File "/venv/lib64/python3.8/site-packages/django/conf/__init__.py", line 82, in __getattr__
    self._setup(name)
  File "/venv/lib64/python3.8/site-packages/django/conf/__init__.py", line 69, in _setup
    self._wrapped = Settings(settings_module)
  File "/venv/lib64/python3.8/site-packages/django/conf/__init__.py", line 170, in __init__
    mod = importlib.import_module(self.SETTINGS_MODULE)
  File "/usr/lib64/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
[...]

Issue #649, opened 2021-08-28
Hierarchy in .secrets.toml overwrites settings in settings.toml


m4rc3l-h3:
Thanks for the great library and your support!

**Describe the bug**
Using hierarchy in .secrets.toml overwrites settings in settings.toml.

**To Reproduce**

1. Having the following folder structure
<details>
<summary> Project structure </summary>

```bash
files/testing/settings.toml
files/testing/.secrets.toml
src/main.py
```
</details>

2. Having the following config files:

<details>
<summary> Config files </summary>

files/testing/settings.toml
```toml
[default]
[default.db]
    name="name_default"
```

and

files/testing/.secrets.toml**
```toml
[default]
[default.db]
    password = 'default_password'
```

</details>

3. Having the following app code:

<details>
<summary> Code </summary>

**/src/main.py**
```python
from dynaconf import LazySettings

settings = LazySettings(
    environments=True,
    ENV_FOR_DYNACONF="default",
    ENVVAR_PREFIX_FOR_DYNACONF="none",
    load_dotenv=True,
    settings_files="../files/testing/settings.toml;../files/testing/.secrets.toml",
)

print(settings.db)
```
</details>

4. Executing under the following environment

<details>
<summary> Execution </summary>

```bash
$ python3 ./src/main.py
```

</details>

**Actual behavior**
Print: {'password': 'default_password'}

**Expected behavior**
Print: {'name': 'name_default', 'password': 'default_password'}

**Environment (please complete the following information):**
 - OS:  Linux 10 (buster)
-  Python: 3.9.1
 - Dynaconf Version: 3.1.5
 - Frameworks in use: none

**Additional context**
Removing hierarchy from .secrets.toml results in expected behavior:

Changing
**files/testing/.secrets.toml**

<details>

```toml
[default]
password = 'default_password'
```
</details>

and

**/src/main.py**

<details>

```python
from dynaconf import LazySettings

settings = LazySettings(
    environments=True,
    ENV_FOR_DYNACONF="default",
    ENVVAR_PREFIX_FOR_DYNACONF="none",
    load_dotenv=True,
    settings_files="../files/testing/settings.toml;../files/testing/.secrets.toml",
)

print(settings.db)
print(settings.password)
```

</details>

Results in the following output:
```bash
{'name': 'name_default'}
default_password
```

m4rc3l-h3:
In addition, when using 
```python
settings.setenv('testing')
```
[...]

PR #951, opened 2023-07-05
Fix - Django functions with `add_converter`

### LINKED ISSUE #949: [bug] CLI not working with Django on master
On master branch the CLI is not working in a Django app

## Steps to Reproduce

```
❯ cd tests_functional/django_example
```

```
❯ DJANGO_SETTINGS_MODULE=foo.settings PYTHONPATH=. dynaconf list
Django app detected


╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮

│                                                                                                  │
│ /home/rochacbruno/Projects/dynaconf/dynaconf/cli.py:546 in _list                                 │
│                                                                                                  │
│   543 │   if env:                                                                                │
│   544 │   │   settings.setenv(env)                                                               │
│   545 │                                                                                          │
│ ❱ 546 │   cur_env = settings.current_env.lower()                                                 │
│   547 │                                                                                          │
│   548 │   if cur_env == "main":                                                                  │
│   549 │   │   flat = True                                                                        │
│                                                                                                  │
│ ╭───────────────────────────────── locals ─────────────────────────────────╮                     │
│ │     _all = False                                                         │                     │
│ │      ctx = <dynaconf.vendor.click.core.Context object at 0x7f4d5a77c130> │                     │
│ │      env = None                                                          │                     │
│ │     flat = False                                                         │                     │
│ │      key = None                                                          │                     │
│ │   loader = None                                                          │                     │
│ │     more = None                                                          │                     │
│ │   output = None                                                          │                     │
│ │ settings = <LazySettings "foo.settings">                                 │                     │
│ ╰──────────────────────────────────────────────────────────────────────────╯                     │
│                                                                                                  │
│ /home/rochacbruno/Projects/dynaconf/.venv/lib/python3.8/site-packages/django/conf/__init__.py:88 │
│ in __getattr__                                                                                   │
│                                                                                                  │
│    85 │   │   """Return the value of a setting and cache it in self.__dict__."""                 │
│    86 │   │   if self._wrapped is empty:                                                         │
│    87 │   │   │   self._setup(name)                                                              │
│ ❱  88 │   │   val = getattr(self._wrapped, name)                                                 │
│    89 │   │                                                                                      │
│    90 │   │   # Special case some settings which require further modification.                   │
│    91 │   │   # This is done here for performance reasons so the modified value is cached.       │
│                                                                                                  │
│ ╭─────────────── locals ───────────────╮                                                         │
│ │ name = 'current_env'                 │                                                         │
│ │ self = <LazySettings "foo.settings"> │                                                         │
│ ╰──────────────────────────────────────╯                                                         │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
AttributeError: 'Settings' object has no attribute 'current_env'
```

Same error happens with `get` command

```
❯ DJANGO_SETTINGS_MODULE=foo.settings PYTHONPATH=. dynaconf get DEBUG


│ /home/rochacbruno/Projects/dynaconf/dynaconf/cli.py:475 in get                                   │
│                                                                                                  │
│   472 │   if default is not empty:                                                               │
│   473 │   │   result = settings.get(key, default)                                                │
│   474 │   else:                                                                                  │
│ ❱ 475 │   │   result = settings[key]  # let the keyerror raises                                  │
│   476 │                                                                                          │
│   477 │   if unparse:                                                                            │
│   478 │   │   result = unparse_conf_data(result)                                                 │
│                                                                                                  │
│ ╭───────────────────────────────── locals ─────────────────────────────────╮                     │
│ │      ctx = <dynaconf.vendor.click.core.Context object at 0x7f32e513c130> │                     │
│ │  default = <dynaconf.utils.functional.Empty object at 0x7f32e4a5d0d0>    │                     │
│ │      env = None                                                          │                     │
│ │      key = 'DEBUG'                                                       │                     │
│ │ settings = <LazySettings "foo.settings">                                 │                     │
│ │  unparse = False                                                         │                     │
│ ╰──────────────────────────────────────────────────────────────────────────╯                     │
│                                                                                                  │
│ /home/rochacbruno/Projects/dynaconf/.venv/lib/python3.8/site-packages/django/utils/functional.py │
│ :259 in inner                                                                                    │
│                                                                                                  │
│   256 │   def inner(self, *args):                                                                │
│   257 │   │   if self._wrapped is empty:                                                         │
│   258 │   │   │   self._setup()                                                                  │
│ ❱ 259 │   │   return func(self._wrapped, *args)                                                  │
│   260 │                                                                                          │
│   261 │   return inner                                                                           │
│   262                                                                                            │
│                                                                                                  │
│ ╭─────────────── locals ───────────────╮                                                         │
│ │ args = ('DEBUG',)                    │                                                         │
│ │ func = <built-in function getitem>   │                                                         │
│ │ self = <LazySettings "foo.settings"> │                                                         │
│ ╰──────────────────────────────────────╯                                                         │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
TypeError: 'Settings' object is not subscriptable

```

And also the new `inspect` command

```
│ /home/rochacbruno/Projects/dynaconf/dynaconf/utils/inspect.py:79 in inspect_settings             │
│                                                                                                  │
│    76 │   original_settings = settings                                                           │
│    77 │   settings = settings if not env else settings.from_env(env)                             │
│    78 │                                                                                          │
│ ❱  79 │   setting_envs = {_env.env for _env in settings._loaded_by_loaders.keys()}               │
│    80 │   if env and env.lower() not in setting_envs:                                            │
│    81 │   │   raise EnvNotFoundError(f"The requested env is not valid: {env!r}")                 │
│    82                                                                                            │
│                                                                                                  │
│ ╭───────────────────── locals ──────────────────────╮                                            │
│ │   ascending_order = True                          │                                            │
│ │     custom_dumper = None                          │                                            │
│ │               env = ''                            │                                            │
│ │   key_dotted_path = 'DEBUG'                       │                                            │
│ │ original_settings = <LazySettings "foo.settings"> │                                            │
│ │     output_format = 'json'                        │                                            │
│ │          settings = <LazySettings "foo.settings"> │                                            │
│ │           to_file = ''                            │                                            │
│ ╰───────────────────────────────────────────────────╯                                            │
│                                                                                                  │
│ /home/rochacbruno/Projects/dynaconf/.venv/lib/python3.8/site-packages/django/conf/__init__.py:88 │
│ in __getattr__                                                                                   │
│                                                                                                  │
│    85 │   │   """Return the value of a setting and cache it in self.__dict__."""                 │
│    86 │   │   if self._wrapped is empty:                                                         │
│    87 │   │   │   self._setup(name)                                                              │
│ ❱  88 │   │   val = getattr(self._wrapped, name)                                                 │
│    89 │   │                                                                                      │
│    90 │   │   # Special case some settings which require further modification.                   │
│    91 │   │   # This is done here for performance reasons so the modified value is cached.       │
│                                                                                                  │
│ ╭─────────────── locals ───────────────╮                                                         │
│ │ name = '_loaded_by_loaders'          │                                                         │
│ │ self = <LazySettings "foo.settings"> │                                                         │
│ ╰──────────────────────────────────────╯                                                         │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
AttributeError: 'Settings' object has no attribute '_loaded_by_loaders'
```

## Investigation

In the django context the settings is being applied correctly

```
❯ DJANGO_SETTINGS_MODULE=foo.settings PYTHONPATH=. django-admin shell
In [1]: from django.conf import settings

In [2]: settings.__class__
Out[2]: dynaconf.base.Settings
```

However I added a breakpoint on line 545 of cli.py and figured out that `settings` there is a bare `Django.conf.settings` not a dynaconf.base.Settings instance, for some reason the dynaconf plugin is not being loaded at that point.

## Expected

At the same directory, checking out to 3.2.12 branch everything works as expected.

```
❯ git checkout 3.1.12

❯ DJANGO_SETTINGS_MODULE=foo.settings PYTHONPATH=. dynaconf list            
Django app detected
Working in production environment 
BASE_DIR<str> '/home/rochacbruno/Projects/dynaconf/tests_functional/django_example'
STATIC_URL<str> '/static/'
...


❯ DJANGO_SETTINGS_MODULE=foo.settings PYTHONPATH=. dynaconf get DEBUG
True%    

## Inspect command does not exist on that version.

❯ DJANGO_SETTINGS_MODULE=foo.settings PYTHONPATH=. django-admin shell
In [1]: from django.conf import settings

In [2]: settings.__class__
Out[2]: dynaconf.base.Settings

```

pedro-psb:
Fixes #949 and provides workaround for #648 

## Overview:
- revert refactor in #939 that caused CLI malfunction
- add workaround to avoid `AppRegistryNotReady` exception when registering Django converters (wrap in special hook)
- unrelated mechanism to turn off recursively evaluation temporarily (needs improvement and isn't decisive to this solution)
- update docs

### Did not test #648
Although this solution looks a good workaround for the issue, I've not included specific test for it because I'm not familiar with `gettext` to think in a good test setup. It should be done later.

### Workaround alternative
My first attempt to solve the issue was wrapping the import for each function, but the new way seems more user friendly.

```python
def reverse_lazy_wrapper(*args, **kwargs):
    import django.urls import reverse_lazy
    return reverse_lazy(*args, **kwargs)

dynaconf.add_converter("reverse_lazy", reverse_lazy_wrapper)
```


tests_functional/django_example/foo/settings.py
```diff
@@ -65,10 +65,16 @@
...
 
-dynaconf.add_converter("reverse_lazy", reverse_lazy)
+
+def django_ready():
+    from django.urls import reverse_lazy  # noqa
+
+    dynaconf.add_converter("reverse_lazy", reverse_lazy)
+
+
+dynaconf.django_ready = django_ready
```
rochacbruno, line 77: We can use a decorator to register this kind of  hook and allow it to be a sequence?

This will require the `django_ready` to be a decorator that registers the functions in a list of hooks like.
`dynaconf/__init__.py`
```py
from functools import wraps

# Key can optionally be an enum here
HOOKS = {
    "django_ready": [],
    ... 
}

def instance_hook(identifier):
    def decorator(f):
        HOOKS[identifier].append(f)
     return decorator
```

Then in the `execute_loaders` 
```py
for _hook in dynaconf.HOOKS["django_ready"]:
    _hook()

# notes
# we may need to avoid executing the same hook twice
#  key can be enum, hook can be a better structured class instance if needed
# I am ok on using strings and just assuming callables.
```

Usage:

```py
import dynaconf

@dynaconf.instance_hook("django_ready")
def add_reverse_url_converter():
    from django.urls import reverse_lazy  # noqa
    dynaconf.add_converter("reverse_lazy", reverse_lazy)

```

pedro-psb:
Yesterday I've experimented a bit with django translation system and I come up with some conclusions.

## Gettext

The django gettext issue #649 is more complex than I thought and won't (don't have to) work with the same solution provided to fix the CLI, so I'll leave it out of this PR.

Just as a note, I've found that `AppConfig` has a `ready` hook that is called when the App registry is loaded, so there should be a good place to put gettext. Still, I need to experiment a little more.

```python
from __future__ import annotations

from django.apps import AppConfig


class PollsConfig(AppConfig):
    name = "foo"

    def ready(self):
        from django.utils.translation import gettext_lazy as _
        from django.conf import settings
        settings.TRANSLATABLE=_("hello world")
```

## CLI fix design

For the actual django fix, I've changed my mind about the design.
- I've realized the proposed hook was redundant as it was called almost at the same spot as the existing (I've tested, it works the same). As it is just a fix, we could avoid creating new machinery and using the existing one.
- Maybe it would be worth adding a shortcut for defining those existing hooks inside `settings.py`, but without creating a new hook system. It may be done as:
  - keyword argument in `DynaconfDjango/Settings`, such as `post_hooks: Callable | [Callable]`
  - `dynaconf.add_post_hook(Callable)`
  - decorator such as `@dynaconf.post_hook`

These are the design, summarized:

```python
# 1) existing hook solution (at dynaconf_hooks.py)

def post():
	...

# 2) add_hook utility for existing loader hooks
def setup_function():
	...

dynaconf.add_hook(setup_function)

# 3) new keyword in DynaconfDjango (and Setting, consequently)
def setup_function():
	...

settings = DynaconfDjango(
	__name__,
	post_hook=setup_function # or [setup_function]
)

# 4) decorator design
@dynaconf.loading_hook
def setup_function():
	...

setup_function() # needs to be called or module-collected

```

What do you think about using the existing hook system instead of creating a new one? If you agree, which design do you prefer? (I go with 3)

rochacbruno:
@pedro-psb I agree with option 3 as it doesn't require adding new concepts.

rochacbruno:
will the hooked function receive the settings as argument? just like the existing dynaconf_hooks?

Issue #926, opened 2023-04-26
dynaconf_merge=False not working with hooks [bug]


Holzhauer:
**Describe the bug**
The documentation (https://www.dynaconf.com/advanced/#hooks) states: "To disable the merging of the data, you can set the dynaconf_merge to False." However, this is not true. Even setting dynaconf_merge to False merges the object returned by post() into existing settings.

**To Reproduce**

1. Having the following folder structure

<details>
<summary> Project structure </summary>

```bash
files/testing/settings.toml
files/testing/dynaconf_hooks.py
src/main.py
```
</details>

2. Having the following config files:

<details>
<summary> Config files </summary>

**/path/.env**
```bash
Your .env content here
```

and

**/path/settings.toml**
```toml
[default]
DEBUG = true
Name="Test"
```

</details>

3. Having the following app code:

<details>
<summary> Code </summary>

**main.py**
```python
from dynaconf import LazySettings

settings = LazySettings(
    environments=True,
    ENV_FOR_DYNACONF="default",
    ENVVAR_PREFIX_FOR_DYNACONF="none",
    load_dotenv=True,
    settings_files="../files/testing/settings.toml",
)

print(settings.as_dict())
```

**dynaconf_hooks.py**
```python
def post(settings):
    data = {"dynaconf_merge": True}
    if settings.DEBUG:
        data["DATABASE_URL"] = "sqlite://"
    return data
```

</details>

4. Executing under the following environment

<details>
<summary> Execution </summary>

```bash
# other commands and details?
# virtualenv activation?

$ python main.py
```

</details>

**Expected behavior**
{'LOAD_DOTENV': True, 'DATABASE_URL': 'sqlite://'}

**Actual behavior**
{'LOAD_DOTENV': True, 'DEBUG': True, 'NAME': 'Test', 'DATABASE_URL': 'sqlite://'}

**Environment (please complete the following information):**
 - OS: Windows 11]
 - Dynaconf Version 3.1.12

pedro-psb:
This is a more general issue. The current implementation of local `dynaconf_merge`s does not support the False flag as it should.
Being tracked in #835

pedro-psb:
In this specific example, it is expected that the values will merge even if merge is set to `False', because top level structure always merges (it isn't affected by the merge flags and marks).

I've added a workaround in #957 functional test for this, which consists in:
- explicit deleting other keys OR
- nesting values in wrapper namespace (which now works with `dynaconf_merge=False`).

[...]

PR #957, opened 2023-07-12
Fix - Enable merge equal False

### LINKED ISSUE #835: dynaconf_merge = false supported?
Hello, I would like to know if it is possible to disable merging for some keys of a configs. In our case, we have a bunch of config files (~30) and merging them is the usual behavior. We do this through the `Dynaconf` parameter `merge_enabled`. Now, in a few cases the merging leads to undesired behavior and it would be nice to turn off merging for these specific config keys. Is it possible to do this easily without changing up the default merge behavior and touch all the config files? I tried `dynaconf_merge = false` as a property but this behavior doesn't seem to be supported.

Example:
`1.toml`
```toml
[general]
a = 1

[other]
a = 2
```

`2.toml`
```toml
[general]
b = 1

[other]
dynaconf_merge = false
b = 2
```

`config.py`
```python
from dynaconf import Dynaconf

settings = Dynaconf(
    settings_files=["1.toml", "2.toml"],
    merge_enabled=True,
)

print(settings.to_dict())
```

Output:
```
{'GENERAL': {'b': 1, 'a': 1}, 'OTHER': {'b': 2, 'a': 2}}
```

Expected Output:
```
{'GENERAL': {'b': 1, 'a': 1}, 'OTHER': {'b': 2}}
```

Thanks for your time.

pedro-psb:
Fix #835 and #926 

## Overview

The problem was that `False` was treated as a general fallback and would not be used even if located in a high override-priority position:
* local-scope marks
* file-scope marks
* global-flag

To solve it, I applied the same strategy used in some places of the codebase of using the `empty` sentinel for not-set values, thus making override priority relationships more explicit.

### DIFF
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -33,7 +33,7 @@ from dynaconf.utils import RENAMED_VARS
 from dynaconf.utils import upperfy
 from dynaconf.utils.boxing import DynaBox
 from dynaconf.utils.files import find_file
-from dynaconf.utils.functional import empty
+from dynaconf.utils.functional import Empty, empty
 from dynaconf.utils.functional import LazyObject
 from dynaconf.utils.parse_conf import apply_converter
 from dynaconf.utils.parse_conf import converters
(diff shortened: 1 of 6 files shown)


rochacbruno (approved):
🎉

PR #958, opened 2023-07-12
Inspect Review

pedro-psb:
This is a review PR for the Inspect feature #939 

## Proposals

(I'm actively updating this section)

### Bug Fixes

- [x] CLI "django app detected" message breaks the output format (e.g. `dynaconf inspect | jq`)
- [x] `pre_load` and `dynaconf_hooks` identifiers are not preserved in history.
- [x] Compatibility break when defining `loader_identifier` with `str` (because of the new `SourceMetadata`)

### API changes in `dynaconf.utils.inspect`

- [x] rename `key_dotted_path`:
from `inspect_settings(key_dotted_path: str, ...)` to `inspect_settings(key: str, ...)`
from `get_history(key_dotted_path: str, ...)` to `get_history(key: str, ...)`

Justification: Explicit names are usually good, but in the context of Dynaconf `key` is used as dotted-path (as in `setting.get` for example), which makes it more consistent. Also, the docstring makes the usage clear and, if we support other path access (like `foo__bar`), `key` can accommodate that semantic.

- [x] merge/rename `custom_dumper` and `output_format` to `dumper: Callable | str (preset string)`
- [x] enforce the use of kwargs except for `settings`
- [x] add `report_builder` to customize the structure of the output report

### `Inspect` default report behavior

- [x] Change `inspect` output
  from `header.filter.history_order=ascdending | descending`
  to `header.chronological_sort=True | False` (more clear information)
- [x] Limit the number of history entries, similar to `git log -1` or `head -n 1`. Use 1 by default (users feedback)

### Clarify History Report behavior

The behavior of the history report may be counter-intuitive, specially with merge flag. It kinda snapshots the state of settings at the given source, but not exactly. This should be better explored, defined and documented.

Example, when source B merges data after A:
```python
# A loads [1, 2, 3], then: settings inspect -k LISTY:
A -> listy=[1, 2, 3]

# B loads listy=[4, "dynaconf_merge"], then: settings inspect -k LISTY:
B -> listy=[1, 2, 3, 4] # not listy=[4]
A -> listy=[1, 2, 3]
```
### Ideas
- Add `cli_inspect_options={...}` to `Dynaconf` instantiation to set custom default parameters (like limit and sort)

### General
- [ ] Add tests for new uncovered inspect-related code
- [x] Update documentation

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -113,7 +113,7 @@ foo: from_yaml
 $ python
 >>> from dynaconf import Dynaconf, inspect_settings
 >>> settings = Dynaconf()
->>> inspect_settings(settings, key_dotted_path="foo", output_format="yaml")
+>>> inspect_settings(settings, key_dotted_path="foo", dumper="yaml")
 header:
   filters:
     env: None
@@ -139,7 +139,7 @@ history:
 To save to a file use:
 
 ```python
-inspect_setting(setting, to_file="filename.json", output_format="json")
+inspect_setting(setting, to_file="filename.json", dumper="json")
 ```
 
 Dynaconf supports some builtin formats, but you can use a custom dumper too that
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -81,7 +81,7 @@ settings = Dynaconf(post_hooks=hook_function)
 You can also set the merging individually for each settings variable as seen on
 [merging](/merging/) documentation.
 
-## `inspect_settings`
+## Inspecting History
 
 > **NEW** in version 3.2.0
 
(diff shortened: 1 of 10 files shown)


rochacbruno:
When dealing with inspect I found some things worth revisiting.

1. The CLI is printing "django app detected" and this breaks piping the output to json parser, we need to omit that line
2. Dynaconf_hooks.py will mutate the settings, but history is not kept, we need to track it, maybe by adding dynamic identifiers on every executed hook function
3. On a django application, when multiple settings.PY files are loaded, the individual filename is not preserved in history
4. When using 	`pre_load` the identifier is also not preserved in history.

I am adding this notes here because I will test more before I add issues

rochacbruno:
@pedro-psb my pr #975 fixed some of the issues listed here, so what is pending is:

- [ ] Limit history results like `git log -n x`
- [ ] Renaming key_dotted_path parameter
- [ ] Output change for ascending/chrono order

If you prefer, we can have separate issues for those.

About the `Clarify History Report behavior` I also changed that on my PR #975 as now the history saves the value pre-parsing, we are going to see in history the state of value before merging.

So we have to keep the header where the `current_value` is displayed, that will be the only place to see the final value.

pedro-psb:
> @pedro-psb my pr #975 fixed some of the issues listed here, so what is pending is:
> 
>     * [ ]  Limit history results like `git log -n x`
> 
>     * [ ]  Renaming key_dotted_path parameter
> 
>     * [ ]  Output change for ascending/chrono order
> 
> 
> If you prefer, we can have separate issues for those.
> 
> About the `Clarify History Report behavior` I also changed that on my PR #975 as now the history saves the value pre-parsing, we are going to see in history the state of value before merging.
> 
> So we have to keep the header where the `current_value` is displayed, that will be the only place to see the final value.

Great! I think there is no need for separate issues. I'll just start working on those remaining tasks and then update the documentation with changes included in the PR, such as the new `--all` capability.

pedro-psb:
@rochacbruno I want an opinion on some changes.

### Output presentation

The general purpose is to make the output interpretation more straightforward and easily definable:
- `header`: contains info on how inspect was called
- `current`: contains the current value as defined in header (filtered key/env)
- `history`: contains the loading history as defined in header (filtered key/env + sorting/limiting)

```yaml
# Original
header:
  filters:
    key: None
    env: None
    history_ordering: "ascending"
  active_value:
    {...}
history:
  [...]

# Proposed
header:
  key_filter: None
  env_filter: None
  new_first: True
  history_limit: None
current:
  {...}
history:
  [...]
```

### Limit history default

Should the history limit be set to 1 or to `all`, by default?

### Limit CLI name

I intent to use this:

```bash
inspect [-n|--limit] <number> # similar to gnu 'head' and 'git log'
```

rochacbruno:
LGTM! @pedro-psb I like having current value separated and the `-n` option is also useful.

We must default for showing everything by default and then allow `-n` to start limiting, just like git log does

rochacbruno (approved):

PR #970, opened 2023-07-27
Doc advanced usage for cli overrides dynaconf settings fix #967

### LINKED ISSUE #967: Using command line arguments to override values in settings file
Hi, Im trying to use command line arguments to override values in settings file something similar to spring boot way to use java properties. Is it possible with dynaconf?

hugoprudente:
Add documentation and `test_functional` for 

- argparse
- click


docs/advanced.md
```diff
@@ -145,6 +145,79 @@ inspect_setting(setting, to_file="filename.json", output_format="json")
...
+from dynaconf import settings
+
+import click
+
+
+@click.command()
+@click.option("--host", default=settings.HOST, help="Host")
+@click.option("--port", default=settings.PORT, help="Port")
+@click.option("--env", default=settings.current_env, help="Env")
+def app(host, port, env):
```
rochacbruno, line 208: I ma curious on how click would handle `def app(**options)` here, then `options` could be passed direct to dynaconf.

hugoprudente: I can test with that too and see how it would handle and add both cases on the test too.

rochacbruno (approved):

docs/advanced.md
hugoprudente: Added the example on the next commit, it works like a charm.

Issue #969, opened 2023-07-27
YAML: Integer Keys Stored as Strings 


sanssecours:
**Describe the bug**

Integer keys in A YAML configuration file are returned as strings, not as integers

**To Reproduce**

1. Having the following folder structure

   <details>

   ```
   .
   ├── config.yaml
   └── test.py

   2 directories, 1 file

   ```

   </details>

2. Having the following config files:

   <details>
   <summary> Config files </summary>

   **config.yaml**

   ```yaml
   tools:
     1: something
     2: moar
   ```

   </details>

3. Having the following app code:

   <details>
   <summary> Code </summary>

   **test.py**

   ```python
   from pathlib import Path

   from dynaconf import Dynaconf

   settings = Dynaconf(settings_files=[Path(__file__).parent / "config.yaml"])

   print(settings.tools)
   ```

   </details>

4. Executing under the following environment

   <details>
   <summary> Execution </summary>

   ```bash
   python test.py
   ```

   </details>

**Expected behavior**

The code should print the following output (and does so using Dynaconf `3.1.12`):

```
{1: 'something', 2: 'moar'}
```

**Actual behavior**

The code prints the following output:

```
{'1': 'something', '2': 'moar'}
```

**Environment:**

- OS: macOS `13.5`
- Dynaconf Version `main` (or `3.2.0`)

**Additional context**

I do not think that this is a problem of [`ruamel.yaml`](https://pypi.org/project/ruamel.yaml/) since the following code:

```py
from pathlib import Path

from ruamel.yaml import YAML

YAML = YAML()
print(YAML.load(Path("config.yaml"))["tools"])
```

prints the expected output:

```
{1: 'something', 2: 'moar'}
```

rochacbruno:
I am not considering this a bug because this is expected and a known caveat of our key system, I am open to change it adding the functionality.

sanssecours:
Thank you for the quick reply.

> I am not considering this a bug because this is expected and a known caveat of our key system, I am open to change it adding the functionality.

That makes sense. Thank you for considering to change the behavior. Just for reference: My use case was code that looks something like this:

```python
if tool_number in settings.tools:
    # do something useful
```

rochacbruno:
So, this is actually a bug introduced on 3.2.0 version, Box keys are being converted to string

## 3.1.12
[...]

PR #975, opened 2023-08-09
Hooking: Add support for changing the wrapped class

rochacbruno:
This PR allows changing the wrapped class with
a class that allows hooks on specific methods.

This will enable https://github.com/ansible/galaxy_ng/pull/1837

This PR also fixes: #969

This PR replaces #828 

Allows to add AFTER and BEFORE hooks to some dynaconf methods

```python
from dynaconf import Dynaconf

def lookup_for_key_in_cache_or_database(settings, value,  key, default):
    return new_value

settings.register_hook(Action.AFTER_GET, lookup_for_key_in_cache_or_database)
```

Then in any part of code

```python
>>> from config import settings
>>> settings.MY_AWESOME_KEY
"doing things to lookup key 'MY_AWESOME_KEY' on cache or db"
```

### DIFF
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -946,14 +946,6 @@ class Settings:
                 validate=validate,
             )
 
-        # Fix for #905
-        # parsed_conf default value was causing duplication
-        # value_not_parsed = (
-        #     value
-        #     if loader_identifier
-        #     and loader_identifier.loader == "validation_default"
-        #     else None
-        # )
         parsed = parse_conf_data(value, tomlfy=tomlfy, box_settings=self)
         key = upperfy(key.strip())
 
@@ -993,13 +985,6 @@ class Settings:
                 source_metadata = source_metadata._replace(merged=True)
                 parsed = object_merge(existing, parsed)
             else:
-                # # Fix for #905
-                # if (
-                #     loader_identifier
-                #     and loader_identifier.loader == "validation_default"
-                # ):
-                #     value = value_not_parsed
-                # else:
                 # `dynaconf_merge` may be used within the key structure
                 # Or merge_enabled is set to True
                 parsed, source_metadata = self._merge_before_set(
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -229,7 +229,7 @@ class Settings:
         self._deleted = set()
         self._store = DynaBox(box_settings=self)
         self._env_cache = {}
-        self._loaded_by_loaders: dict[SourceMetadata, Any] = {}
+        self._loaded_by_loaders: dict[SourceMetadata | str, Any] = {}
         self._loaders = []
         self._defaults = DynaBox(box_settings=self)
         self.environ = os.environ
@@ -251,9 +251,13 @@ class Settings:
 
         compat_kwargs(kwargs)
         if settings_module:
-            self.set("SETTINGS_FILE_FOR_DYNACONF", settings_module)
+            self.set(
+                "SETTINGS_FILE_FOR_DYNACONF",
+                settings_module,
+                loader_identifier="init_settings_module",
+            )
         for key, value in kwargs.items():
-            self.set(key, value)
+            self.set(key, value, loader_identifier="init_kwargs")
         # execute loaders only after setting defaults got from kwargs
         self._defaults = kwargs
 
@@ -385,9 +389,7 @@ class Settings:
                 )
             )
         )
-        loader_identifier = SourceMetadata(
-            "validation_default", "unique", env.lower()
-        )
+        loader_identifier = SourceMetadata("setdefault", "unique", env.lower())
 
         if apply_default:
             self.set(
@@ -894,7 +896,7 @@ class Settings:
         self,
         key,
         value,
-        loader_identifier: SourceMetadata | None = None,
+        loader_identifier: SourceMetadata | str | None = None,
         tomlfy=False,
         dotted_lookup=empty,
         is_secret="DeprecatedArgument",  # noqa
@@ -910,6 +912,13 @@ class Settings:
         :param merge: Bool define if existing nested data will be merged.
         :param validate: Bool define if validation will be triggered
         """
+        if isinstance(loader_identifier, str):
+            loader_identifier = SourceMetadata(
+                loader="undefined",
+                identifier=loader_identifier,
+                merged=merge is True,
+            )
+
         if validate is empty:
             validate = self.get("VALIDATE_ON_UPDATE_FOR_DYNACONF")
 
@@ -1292,7 +1301,11 @@ class Settings:
 
         if self._loaded_files:  # called once
             root_path = os.path.dirname(self._loaded_files[0])
-            self.set("ROOT_PATH_FOR_DYNACONF", root_path)
+            self.set(
+                "ROOT_PATH_FOR_DYNACONF",
+                root_path,
+                loader_identifier="_root_path",
+            )
             return root_path
 
     def load_extra_yaml(self, env, silent, key):
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -765,7 +765,11 @@ class Settings:
             box_settings=self,
         )
         if settings_module != getattr(self, "SETTINGS_MODULE", None):
-            self.set("SETTINGS_MODULE", settings_module)
+            self.set(
+                "SETTINGS_MODULE",
+                settings_module,
+                loader_identifier="settings_module_method",
+            )
 
         # This is for backewards compatibility, to be removed on 4.x.x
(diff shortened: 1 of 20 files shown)


dynaconf/cli.py
```diff
@@ -50,7 +50,7 @@ def set_settings(ctx, instance=None):
 
     settings = None
 
-    _echo_enabled = ctx.invoked_subcommand not in ["get", None]
+    _echo_enabled = ctx.invoked_subcommand not in ["get", "inspect", None]
```
rochacbruno, line 53: @pedro-psb this fixed the inspect `django app detected` problem

dynaconf/utils/inspect.py
```diff
@@ -196,6 +200,16 @@ def get_history(
...
         result.append({**source_metadata._asdict(), "value": data})
+
+    if not result:
+        # Key may be set in obj but history not tracked
+        if (data := obj.get(key_dotted_path, empty)) is not empty:
+            generic_source_metadata = SourceMetadata(
+                loader="undefined",
+                identifier="undefined",
+            )
+            result.append({**generic_source_metadata._asdict(), "value": data})
```
rochacbruno, line 211: @pedro-psb I added this because there will be keys without tracking history, if someone implemented a custom loader and is not passing an identifier, this way it still shows on the `inspect` output as `undefined` as we cannot track the history.

tests_functional/issues/969/app.py
```diff
@@ -0,0 +1,12 @@
...
+from dynaconf import Dynaconf
+
+data = {1: 2}
+settings = Dynaconf()
+
+# __import__('ipdb').set_trace()
+settings.set("data", data)
+
+
+assert settings["data"][1] == 2
```
rochacbruno, line 12: This is to ensure issue #969

dynaconf/utils/inspect.py
pedro-psb: That's a good improvement.
I think there should be a `data=_ensure_serializable(data)` to sanitize the data (for the dumpers) before appending to `result`, as done in the previous history-generating loop.

dynaconf/base.py
```diff
@@ -248,9 +251,13 @@ def __init__(self, settings_module=None, **kwargs):  # pragma: no cover
...
         if settings_module:
-            self.set("SETTINGS_FILE_FOR_DYNACONF", settings_module)
+            self.set(
+                "SETTINGS_FILE_FOR_DYNACONF",
+                settings_module,
+                loader_identifier="init_settings_module",
+            )
         for key, value in kwargs.items():
-            self.set(key, value)
+            self.set(key, value, loader_identifier="init_kwargs")
```
rochacbruno, line 260: @pedro-psb I added tracking history for when `Dynaconf(foo="bar")` parameters are passed and set to the object

dynaconf/base.py
```diff
@@ -372,9 +389,7 @@ def setdefault(
                 )
             )
         )
-        loader_identifier = SourceMetadata(
-            "validation_default", "unique", env.lower()
-        )
+        loader_identifier = SourceMetadata("setdefault", "unique", env.lower())
```
rochacbruno, line 392: I changed the name to `setdefault` because this method can be used generally, not exclusive to validators

dynaconf/base.py
```diff
@@ -897,6 +912,13 @@ def set(
         :param merge: Bool define if existing nested data will be merged.
         :param validate: Bool define if validation will be triggered
         """
+        if isinstance(loader_identifier, str):
```
rochacbruno, line 915: If set receives a loader_identifier as str we create a SourceMetadata instance internally, preventing the bug reported by @jperras on https://github.com/fictivekin/dynaconf-aws-loader/pull/15 

c/c @pedro-psb

dynaconf/cli.py
```diff
@@ -834,7 +834,15 @@ def validate(path):  # pragma: no cover
...
-def inspect(key, env, format, descending):  # pragma: no cover
+@click.option(
+    "--all",
+    "_all",
+    "-a",
+    default=False,
+    is_flag=True,
+    help="show dynaconf internal settings?",
+)
+def inspect(key, env, format, descending, _all):  # pragma: no cover
```
rochacbruno, line 845: I added an option to show all internal values when running `dynaconf inspect` without key

dynaconf/loaders/__init__.py
```diff
@@ -39,7 +39,7 @@ def default_loader(obj, defaults=None):
     for key in all_keys:
         if not obj.exists(key):
             value = defaults.get(key, default_settings_values.get(key))
-            obj.set(key, value)
+            obj.set(key, value, loader_identifier="default_settings")
```
rochacbruno, line 42: @pedro-psb I added tracking for default settings loaded from `default_settings.py` and envvars, by default omited from inspect, but passing --all it will be visible

dynaconf/loaders/__init__.py
```diff
@@ -169,13 +174,25 @@ def _run_hook_function(
 
     # update obj settings
     if hook_dict:
+        identifier = f"{hook_func.__name__}@{hook_source}"
```
rochacbruno, line 177: I added tracking also for `dynaconf_hooks.py` loaded data

dynaconf/utils/inspect.py
```diff
@@ -177,12 +185,24 @@ def get_history(
             }
         ]
     """
+    # trigger key based hooks
+    if key_dotted_path:
+        obj.get(key_dotted_path)  # noqa
```
rochacbruno, line 190: This is required because if value is dynamic, loaded only on an AFTER_GET hook, it is needed first to get it so the tracking is added to history

tests_functional/issues/905_item_duplication_in_list/app.py
```diff
@@ -18,6 +18,6 @@
 
 settings.validators.validate()
 
-assert settings.group.test_list == ["'1'", "'2'"]
+assert settings.group.test_list == ["1", "2"], settings.group
```
rochacbruno, line 21: I don't know why this test was passing before, but if envvar is `["'1'", "'2'"]` the Python result must be `["1", "2"]`

dynaconf/base.py
```diff
@@ -891,124 +910,120 @@ def set(
...
-        )
-
-        value = parse_conf_data(value, tomlfy=tomlfy, box_settings=self)
+        # value_not_parsed = (
+        #     value
+        #     if loader_identifier
+        #     and loader_identifier.loader == "validation_default"
+        #     else None
+        # )
+        parsed = parse_conf_data(value, tomlfy=tomlfy, box_settings=self)
```
pedro-psb, line 957: This looks way better

dynaconf/hooking.py
```diff
@@ -0,0 +1,305 @@
...
+
+        function_name = name or fun.__name__
+
+        # function being called not in the list of hooks, return the original
+        if not set(_registered_hooks).intersection(
+            (f"before_{function_name}", f"after_{function_name}")
+        ):
+            return fun(self, *args, **kwargs)
+
+        # Create an unhookable (to avoid recursion)
```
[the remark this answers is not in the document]
rochacbruno: @pedro-psb yes, ideally the same object would pass down, however if the same key is evaluated there this ends in an infinite loop.

There is still an infinite loop problem I am trying to solve, but this one is not common.

rochacbruno:
@pedro-psb this PR started to just add the Hookable support but at the end I needed to refactor things related to `inspect`.

I think it is in a mergeable state, so lets review and merge this before it became too big, so it can unblock all the other inspect review you have in progress (sorry but there will be merge conflicts on that one)

rochacbruno:
Summary of changes:

- add support for _wrapper_class replacement
- add hooking with a Hookable implementation to be used by Django integration only (this is transparent now, must not affect any current user)
- Changes on inspect
    - History now saves the raw value instead of parsed, as parsed will be showed on the current_value anyway
    - Set will always create a source_metadata even if not passed or if a str is passed as laoder_identifier
    - Internal uses of set are now tracked
    - dynaconf_hooks.py is now tracked
    - set method now keeps `value` as the raw unparsed value and `parsed` as the evaluated value
    - `inspect` command no more prints `django app detected`
    - get_history now accepts `include_internal` to output internal identifiers
    - `_get_data_by_key` refactored for dotted path traversal
    - access to integer indexes on dotted path removed, we gonna implement later as `key._1_`
    - Fixed the acceptance of keys with other types

dynaconf/utils/inspect.py
pedro-psb, line 1: I don't remember exactly why I added this option, so I guess we're fine without it

dynaconf/base.py
jperras: @rochacbruno oh awesome! Thanks for this!

PR #978, opened 2023-08-14
Allow json caster to parse boolean correctly

EdwardCuiPeacock:
About a year ago, I made a contribution to add a feature to allow type casting (https://github.com/dynaconf/dynaconf/pull/704). In my more recent use case, I wanted to use `@json` to get a dictionary that contains boolean values, like the following:

```yaml
my_field: 
  key: 
    attribute: True 

field_name: my_field

value: "@json @jinja {{ (this|attr(this.field_name)).get('key') or {} }}"
```

However, this throws the following error:

```
lambda x: json.loads(x.replace("'", '"'))
  File "xxxx/lib/python3.9/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "xxxx/lib/python3.9/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "xxxx/lib/python3.9/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 129 (char 128)
```

Which suggests to me that the json string parsing was not able to handle boolean values correctly. The json string to be parsed are in the form of

```
"{'attribute': True}"
```
which `json.loads` cannot handle. It necessarily needs to be the lower case `true` for `json.loads` to recognize this variable as a boolean.

The change I made allows `@json` caster to correctly parse the boolean variables by looking for strings "True" and "False" (without the quote; if it is quoted, then it stays as strings), and replace them with the lower case form.


dynaconf/utils/parse_conf.py
```diff
@@ -228,6 +228,17 @@ def evaluate(settings, *args, **kwargs):
...
 
+def _parse_json_strings(x):
+    x = x.replace("'", '"')  # replace single with double quotes
+    # replace unquoted True / False with lower case true / false
+    if "True" in x:
+        x = re.sub(r'(?<!")\bTrue\b', "true", x)
+    if "False" in x:
+        x = re.sub(r'(?<!")\bFalse\b', "false", x)
+
+    return json.loads(x)
```
rochacbruno, line 239: What would happen in this corner case ?

```python
VALUE = '@json {"TrueKey": False}'
```

pedro-psb: "TrueKey" won't be replaced, but I think it is worth adding this case to the tests

pedro-psb (changes requested):
My suggestions are:
- rename `_parse_json_string` param `x` to `value`
- add a test to assert the replace won't affect undesired `True` and `False` occurrences. e.g
```python
'{"KeyTrue": True}' # ok, will keep "KeyTrue"
'{"Key-True": True}' # fail, will change to "Key-true"
```

dynaconf/utils/parse_conf.py
rochacbruno: do we really need regex here?

what if a simple equality check?

```py
if value in ["True", "False"]:
    value = value.lower()
```

pedro-psb: That doesn't work because value is the whole serialized `'{"TrueKey": False}'` string.

EdwardCuiPeacock: You know, the simple quality makes more sense. I tested this a little bit. The string to parse is probably almost always converted from a dict after parsed by jinja. For example, 

```python
value = str({"field": True})
```

 gives me

```
"{'field': True}"
```

So like what you have above, I can potentially just match the substring `": True`, and `": False`, and replace them with `": true`, and `": false`.  I will work on this.

EdwardCuiPeacock:
> My suggestions are:
> 
> * rename `_parse_json_string` param `x` to `value`
> * add a test to assert the replace won't affect undesired `True` and `False` occurrences. e.g
> 
> ```python
> '{"KeyTrue": True}' # ok, will keep "KeyTrue"
> '{"Key-True": True}' # fail, will change to "Key-true"
> ```

Will do. Will add the test cases.

dynaconf/utils/parse_conf.py
rochacbruno: In this case I think the regex may be a good idea

dynaconf/utils/parse_conf.py
```diff
@@ -239,6 +239,17 @@ def evaluate(settings, *args, **kwargs):
     return evaluate
 
 
+def _parse_json_strings(value):
+    value = value.replace("'", '"')  # replace single with double quotes
```
rochacbruno, line 243: 🚩

What will happen with a JSON like:

```python
{"somekey": "This is a 'value' containing single' quotes"}
```

EdwardCuiPeacock: Sadly, I can't come up with a good regex to replace all the patterns I have considered, 

```python
input_strings =[
    '''{"somekey": "This is a 'value' containing single' quotes", "someotherkey": "Another 'value'"}''',
    """{'somekey': "This is a 'value' containing single' quotes", 'someotherkey': "Another 'value'"}""",
    #"{'somekey': 'This is a 'value' containing single' quotes'}", # this should never happend when converting from dict to str 
    """{'somekey': "This is a 'value' containing single' quotes", 'someotherkey': 'Another value'}""",
    """{'somekey': 'This is a value not containing single quotes', 'someotherkey': 'Another value'}""",
    # """{"somekey": 'This is a 'value',  'containing single' quotes'}""",
]
```

Then how about using `ast.literal_eval`? This would also fix the issue with True / False parsing.

dynaconf/utils/parse_conf.py
```diff
@@ -254,11 +255,9 @@ def evaluate(settings, *args, **kwargs):
     )
     if isinstance(value, Lazy)
     else str(value).lower() in true_values,
-    "@json": lambda value: value.set_casting(
-        lambda x: json.loads(x.replace("'", '"'))
-    )
+    "@json": lambda value: value.set_casting(ast.literal_eval)
     if isinstance(value, Lazy)
-    else json.loads(value),
+    else ast.literal_eval(value),
```
rochacbruno, line 260: I don't think this will address all the cases, look:

Case 1 OK

```py
>>> ast.literal_eval("{'foo': True}")
{'foo': True}
```

Case 2 - NOK

```py
>>> ast.literal_eval("{'foo': true}")
ValueError: malformed node or string on line 1: <ast.Name object at 0x7f4e111cb6d0>
```

It is really trick that we need to make both `true` (valid json) and `True` (valid python) working.

I am rethinking this issue

Should we allow invalid json  as  '{"foo": True}' to be passed to `@json` ?

Maybe we want to make it strict, and add a new marker `@dict` to accept python dict literals.

Another option is adding a filter to Jinja to address your use case @EdwardCuiPeacock 

```
"@json @jinja {{this.FOO | as_bool }}"
```

rochacbruno: We can instead of `dict` add a `@py_literal`

EdwardCuiPeacock: Yep, I didn't consider that case. The tricky part is, when first rendered with @jinja, the dictionary will be converted to strings, probably by `str(value)`, before being further casted by @json. I guess this is why we have `True` instead of `true`, or single quotes `'` instead of double quotes `"`. I am not familiar with the code enough to pinpoint where this string conversion is (maybe under `Lazy._dynaconf_encode`?). But if we can make special cases when converting Python `dict` to strings using `json.dumps` instead of `str`, that may also solve the problem.

rochacbruno: @EdwardCuiPeacock there is another option, custom JSONDEcoder

```py
In [37]: class BoolDecoder(json.JSONDecoder):
    ...:     def raw_decode(self, s, idx=0):
    ...:         try:
    ...:             return super().raw_decode(s, idx=idx)
    ...:         except json.JSONDecodeError:
    ...:             # Handle the replacement here
    ...:             # find :True : True, :False, : False
    ...:             # pattern and replace with lowercase
                     # BETTER USE A REGEX HERE
    ...:             if "True" in s:
    ...:                 s = s.replace("True", "true")
    ...:             if "False" in s:
    ...:                 s = s.replace("False", "false")
    ...:             return super().raw_decode(s, idx=idx)
    ...: 

In [38]: json.loads('{"foo": True, "a": 1}', cls=BoolDecoder)
Out[38]: {'foo': True, 'a': 1}

In [39]: json.loads('{"foo": False, "a": 1}', cls=BoolDecoder)
Out[39]: {'foo': False, 'a': 1}

In [40]: json.loads('{"foo": false, "a": 1}', cls=BoolDecoder)
Out[40]: {'foo': False, 'a': 1}

In [41]: json.loads('{"foo": true, "a": 1}', cls=BoolDecoder)
Out[41]: {'foo': True, 'a': 1}

```

EdwardCuiPeacock:
@rochacbruno I think the casting idea solved my current issue. But here is what I have so far after the update:
* Adding `@py_literal` casting to handle non-json cases
* Added some unit test to test the new caster
* For `@json` casting, remove the hack to replace single quote `'` with double quotes, `"`, as this won't work for cases like `{"somekey": "This is a 'value' containing single' quotes"}`.
* For `@json` casting, adding unit testing which will correct raise `json.decoder.JSONDecodeError` if the input is not strictly json

dynaconf/utils/parse_conf.py
```diff
@@ -254,11 +255,12 @@ def evaluate(settings, *args, **kwargs):
     )
     if isinstance(value, Lazy)
     else str(value).lower() in true_values,
-    "@json": lambda value: value.set_casting(
-        lambda x: json.loads(x.replace("'", '"'))
```
rochacbruno, line 258: I like this change, to be more strict, however this is a breaking change and we will merge this only for 4.0.0

dynaconf/utils/parse_conf.py
```diff
@@ -254,11 +255,12 @@ def evaluate(settings, *args, **kwargs):
...
-    )
+    "@json": lambda value: value.set_casting(json.loads)
     if isinstance(value, Lazy)
     else json.loads(value),
+    "@py_literal": lambda value: value.set_casting(ast.literal_eval)
```
rochacbruno, line 261: Maybe we can just short it to `@pyliteral`

tests/test_utils.py
```diff
@@ -162,33 +162,104 @@ def test_casting_bool(settings):
...
+    )
+    assert isinstance(res, dict)
+    assert "FOO" in res and res["FOO"] == "bar"
+    assert "key" in res and res["key"] is False
+    assert "somekey" in res and "'value'" in res["somekey"]
+
+    # Testing list
+    res = parse_conf_data("""@py_literal ["a", "b", 'c', 1]""")
+    assert isinstance(res, list)
+
```
rochacbruno, line 261: It would be nice to add a test case for `@pyliteral @jinja ...` combination

rochacbruno (dismissed):
LGTM,

However we might keep this PR unmerged until we are ready for 4.0.0 as it includes a breking change.

EdwardCuiPeacock:
@rochacbruno Sounds good to me. Let me know when that happens.

rochacbruno:
BTW we need to document this

 casting idea ` | to_bool` solved my current issue. In fact, i can use `| tojson` within the jinja template to convert to a json string first, then `@json` will correctly parse the rendered string.

```yaml
y_field: 
  key: 
    attribute: True 

field_name: my_field

value: "@json @jinja {{ (this|attr(this.field_name)).get('key') | tojson or {} }}"

```

EdwardCuiPeacock:
Added the new features in docs with examples.

PR #980, opened 2023-08-16
Fix #979 - validators must not parse with toml

### LINKED ISSUE #979: [bug] when using a validator with a default for nested data it parses value via toml twice
**Describe the bug**

```py
from __future__ import annotations

from dynaconf import Dynaconf

settings = Dynaconf()
settings.validators.register(
    Validator("group.something_new", default=5),
)

assert settings.group.test_list == ["1", "2"], settings.group
```

Execution

```console
$ DYNACONF_GROUP__TEST_LIST="['1','2']" python app.py
```

expectation:

```py
settings.group.test_list == ["1", "2"]
```

current behavior

```console
$ DYNACONF_GROUP__TEST_LIST="['1','2']" python app.py
  File "/home/rochacbruno/Projects/dynaconf/tests_functional/issues/905_item_duplication_in_list/app.py", line 21, in <module>
    assert settings.group.test_list == ["1", "2"], settings.group
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: {'something_new': 5, 'TEST_LIST': [1, 2]}
```

## What is happening?

At the end of validators.validate.validate the `Validator("group.something_new", default=5),` is executed and then default value is merged with current data, during that process dynaconf is calling `parse_conf_data` passing `tomlfy=True` which forces toml to evaluate the existing data again so.

```py
In [1]: from dynaconf.utils.parse_conf import parse_conf_data

In [2]: data = {"TEST_LIST": ["1", "2"]}

In [3]: parse_conf_data(data, tomlfy=True)
Out[3]: {'TEST_LIST': [1, 2]}

In [4]: parse_conf_data(data, tomlfy=False)
Out[4]: {'TEST_LIST': ['1', '2']}
```

## Possible solutions:

Change `setdefault` method to use `tomlfy=False` 
Change `parse_with_toml` to avoid doing that transformation.

rochacbruno:
#979 #905
3 years ago I made the wrong decision to pass validators defaults to TOML parser.

Validator object is created at Python level, or at YAML level where there is no need to use toml as parser.

TOML parser must ideally be used only on environment variables where there is no type system available.

That said, this PR ensures toml will keep parsing values coming from envvars, but setdefault method will stop passing `tomlfy=True` so the type informed on the validator will be the type set as default value.

The only possible breaking change on this is if someone is setting defaults relying on TOML to parse the data, in that case, user will need to manually call `parse_with_toml` on default data.

I think that is not a common case, if someone reports it as a broken change, we can provide a new argument `parse_default_with_toml` on the Validator object later.

### DIFF
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -396,7 +396,7 @@ class Settings:
                 item,
                 default,
                 loader_identifier=loader_identifier,
-                tomlfy=True,
+                tomlfy=False,
             )
             return default
 
--- a/dynaconf/validator.py
+++ b/dynaconf/validator.py
@@ -262,7 +262,8 @@ class Validator:
                 and self.is_type_of is str
             ):
                 # avoid TOML from parsing "+-1" as integer
-                default_value = f"'{default_value}'"
+                # by forcing casting to str
+                default_value = f"@str {default_value}"
 
             value = settings.setdefault(
                 name,
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -745,13 +745,13 @@ def test_toml_should_not_change_validator_type_with_is_type_set():
 
 def test_toml_should_not_change_validator_type_with_is_type_not_set_int():
     settings = Dynaconf(
-        validators=[Validator("TEST", default="+172800")]
+        validators=[Validator("TEST", default=-172800)]
         # The ways to force a string is
         # passing is_type_of=str
-        # or default="@str +172800" or default="'+172800'"
+        # or default="@str +172800" or default="+172800"
     )
 
-    assert settings.test == +172800
+    assert settings.test == -172800
 
 
 def test_toml_should_not_change_validator_type_using_at_sign():
--- a/tests_functional/issues/905_item_duplication_in_list/.env
+++ b/tests_functional/issues/905_item_duplication_in_list/.env
@@ -1,3 +1,3 @@
-DYNACONF_GROUP__TEST_LIST = ["'1'", "'2'"]   # double quotes to cast as string
+DYNACONF_GROUP__TEST_LIST="['1', '2']"
 
 DYNACONF_GROUP__ANOTHER = 45
--- a/tests_functional/issues/905_item_duplication_in_list/app.py
+++ b/tests_functional/issues/905_item_duplication_in_list/app.py
@@ -11,7 +11,6 @@ settings = Dynaconf(
     merge_enabled=True,
 )
 
-
 settings.validators.register(
     Validator("group.something_new", default=5),
 )


dynaconf/base.py
```diff
@@ -396,7 +396,7 @@ def setdefault(
                 item,
                 default,
                 loader_identifier=loader_identifier,
-                tomlfy=True,
+                tomlfy=False,
```
rochacbruno, line 399: all other `set` methods such as `set` and `update` have tomlfy=False as a default, the reason for it is for loaders to be able to enable it on-demand, env_loader will always pass `True to that parameter, while other loaders that can rely on a type system will probably not need to parse data again. (but looks like that is not the current case)

If I was going to reimplement this today, I would accept just the bare value and do not parse it, require that the caller parses data if needed, but now it is too late for this change, so lets deal with it.

dynaconf/validator.py
```diff
@@ -262,7 +262,8 @@ def _validate_items(
                 and self.is_type_of is str
             ):
                 # avoid TOML from parsing "+-1" as integer
-                default_value = f"'{default_value}'"
+                # by forcing casting to str
+                default_value = f"@str {default_value}"
```
rochacbruno, line 266: `toml` parsing will not be executed by setdefault, however the converters like `@str` will keep being processed! (this must have been the solution 3 years ago)

tests/test_validators.py
```diff
@@ -745,13 +745,13 @@ def test_toml_should_not_change_validator_type_with_is_type_set():
 
 def test_toml_should_not_change_validator_type_with_is_type_not_set_int():
     settings = Dynaconf(
-        validators=[Validator("TEST", default="+172800")]
+        validators=[Validator("TEST", default=-172800)]
```
rochacbruno, line 748: This is a possible breaking change, but I guess no one is relying on it, it was odd to pass a `str` to a default and expect an integer back.

Now the behavior is more consistent, if default needs to be integer, the user must pass integer or add `@int` as prefix.

tests_functional/issues/905_item_duplication_in_list/.env
```diff
@@ -1,3 +1,3 @@
-DYNACONF_GROUP__TEST_LIST = ["'1'", "'2'"]   # double quotes to cast as string
+DYNACONF_GROUP__TEST_LIST="['1', '2']"
```
rochacbruno, line 1: Double quotes is required to force str internal values, but the recommended way is to pass it outside not internally.
```console
$ DYNACONF_LIST="['1','2']" dynaconf list -k LIST
LIST<BoxList> <BoxList: ['1', '2']>
```

pedro-psb (approved):
I've found tomlfy confusing sometimes, but now its clearer what it purpose was.

I think it is ok to do this break changes as it is low risk and is a design fix. It may be worth to review the documentation on this to assure it is up-to-date (I can do it tomorrow). I'd also suggest adding this to the next minor release (`3.3.0`) and add a upcoming Breaking Change Warning on the changelog of the next patches (`3.2.x`).

pedro-psb:
@rochacbruno This section of the docs needs to be updated:
https://github.com/dynaconf/dynaconf/blob/297eefc1aa8cd097d296d39e48fda32ddb3de68c/docs/validation.md?plain=1#L533-L550

pedro-psb:
And it would be good to add the motivation for `tomlfy` on `parse_conf_data` docstring, it may be usefull for someone digging in the codebase.

rochacbruno:
This implementation will be obsolete and replaced by #981

PR #983, opened 2023-08-21
Fix #976 from envvars parse True/False as booleans

### LINKED ISSUE #976: [RFC] parse True/False as boolean by default
Usually when exporting env vars, users must export booleans as `true/false` lowercase for dynaconf to parse it using toml and return a bool type, however it is a common mistake to do the following:

```
export PREFIX_DEBUG=False
# instead of
export PREFIX_DEBUG=false
```

The env var will be read as `"False"` (str) and toml will parse as a string.

As this is very common, I suggest Dynaconf to transform `True,False` to lowercase before passing to the toml parser, but only if the value is strictly `in ["True", "False"]` and for sake of compatibility problens add an opt-out variable.

rochacbruno:
Fix #976 

Only when reading from envvars True and False will be transformed to lowercase to allow toml parser.

This will work only for bare `True` and `False` as in

```
export DYNACONF_THING=True
```

This will not apply for toml datatypes like

```
export DYNACONF_LIST=["True", "False"]
```


pedro-psb (approved):
I like the approach of making this casting only for envvar loading, as the demand was specific to this case. It is safer than doing a general casting on "True/False" strings.

The solution is straightforward and seems well covered for me. If we spot a blind spot we can fix it later. (the amount of changes needed for such a simple change gave me a refactor idea: move set/update out of loaders, similar to what you'll be doing on validators)

I think it could be shipped along with #980 (3.3.0) with breaking change warning on (3.2.x) (I'm not sure if I'm being too conservative about this, but looks the right thing to do)

docs/envvars.md
```diff
@@ -33,6 +33,10 @@ export DYNACONF_STRING_NUM="'76'"                            # str: "76"
 export DYNACONF_PERSON__IS_ADMIN=true                        # bool: True (nested)
 ```
 
+!!! info 
+    Dynaconf will automatically transform `True` and `False` to `true` and `false` respectively,
+    this transformation allows TOML to parse the value as a boolean. If you want to keep the
```
pedro-psb, line 38: May be worth mentioning that it won't apply for `export DYNACONF_LIST=["True", "False"]`

Issue #984, opened 2023-08-21
[bug] KeyError: '_bypass_evaluation' in Dynaconf 3.2.1


JacobCallahan:
**Describe the bug**
The change introduced in PR #966 also introduced a KeyError in our tests.

**To Reproduce**
Not currently clear without digging further into the cause, but this was triggered in our tests using multi-threading. You can see the failures in the github action link here: https://github.com/SatelliteQE/broker/actions/runs/5928071069/job/16074442428?pr=227

**Expected behavior**
This KeyError should not be triggered. Instead, there should likely be a check that the key exists before attempting to pop or give a default pop value of None.
```python
self._store._box_config.pop("_bypass_evaluation", None)
```

JacobCallahan:
@pedro-psb this should be an easy fix :)

rochacbruno:
@JacobCallahan https://github.com/dynaconf/dynaconf/pull/985 this is a hotfix

However looks like we have a bigger problem with the multi-threading to investigate

JacobCallahan:
Thanks for the quick fix @rochacbruno!

PR #985, opened 2023-08-21
Hotfix bypass evaluation #984

rochacbruno:
Fix https://github.com/dynaconf/dynaconf/issues/984

NOTE: Looks like there is an issue with multithreading
because the state _bypass_evaluation is being mutated on
shared settings obj.


pedro-psb (approved):
I need to review the usage of `_bypass_evaluation`. The intention of this was to reduce unecessary evaluation calls, but as far as I remember it was not critical to the bugfixes  (where it was introduced)

JacobCallahan (approved):
Not a reviewer on this repo, but the change looks good to me.

PR #988, opened 2023-08-23
Stop setting defaults on validators

### LINKED ISSUE #981: [RFC] Stop setting default value inside validate method
There is a problem with the current way of setting default values when using validators.

The method `_validate_items` should have no `side effect` on the object being validated, inside that method, teh `settings` must be read only to perform the validation.

Currently this method calls set twice to set default values and apply cast transformation.

[here](https://github.com/dynaconf/dynaconf/blob/9ca687a4e6df9d67cf060a1fe092a782a256ec24/dynaconf/validator.py#L267) and [here](https://github.com/dynaconf/dynaconf/blob/9ca687a4e6df9d67cf060a1fe092a782a256ec24/dynaconf/validator.py#L295)

This is causing problems with the `inspect` history and is duplicating the calls to `set`

## Proposal

1. Stop calling ` set` on `validate_items`
2. Collect defaults and cast from registered validators

## Refactoring

1. Stop calling `set` and `setdefault` inside `_validate_items`
3. change this [block](https://github.com/dynaconf/dynaconf/blob/9ca687a4e6df9d67cf060a1fe092a782a256ec24/dynaconf/validator.py#L267-L272) to `value = settings.get(name, empty)`
4. On `ValidatorList.register` iterate all validators and register its `default` and `cast` parameter on `settings._validators_defaults: DynaBox` (OR store it on the instance of ValidatorList itself
5. On `Settings.get` [line](https://github.com/dynaconf/dynaconf/blob/9ca687a4e6df9d67cf060a1fe092a782a256ec24/dynaconf/base.py#L521) change the implementation to raise `KeyError` and then on except lookup the `self._validators_defaults` else `default`
6. On [line](https://github.com/dynaconf/dynaconf/blob/9ca687a4e6df9d67cf060a1fe092a782a256ec24/dynaconf/base.py#L522) lookup for `cast` also on `self._validators_defaults` if not passed directly to `get`

rochacbruno:
Default will be declared on validators
but applied only on get method

Fix #981

- [ ] pass tests
- [ ] make default value loading lazy
- [ ] update docs
- [ ] test cast with lazy values


rochacbruno:
copilot:poem

pedro-psb:
@rochacbruno The breaking change in #980 (such as turning off toml parsing on setdefault, or other) will be somehow implemented here? So we can add a warning on the `3.2.2` release note.

eirnym:
Hi @rochacbruno Is there any rough estimate for this PR?

rochacbruno:
Replaced because of #1107 further improvements tracked on https://github.com/dynaconf/dynaconf/issues/1082#issuecomment-2199703596

PR #1031, opened 2023-12-06
replace dead link flask subclassing page

akjems:
The link http://flask.pocoo.org/docs/0.12/patterns/subclassing/ leads to a page not found. I replaced it with the current link.


pedro-psb (approved):
@akjems Thanks for the contribution!

We should really include a link-checker on CI to run every now and then.

PR #1043, opened 2024-01-20
chore(ci): move release workflow to GitHub actions (partial)

pedro-psb:
This replaces #1028 

I've decided to simplify my approach to the CI release workflow:

- The release workflow won't be triggered on tagged commit, only manually.
  - It its safer if we need to do a manual release
- The release assumes what is on master already passed the whole CI tests suite, so those won't be triggered on release
  - only build/sanity checks will be performed, as the release script doesn't change relevant lib code.
  - I believe the tests are triggered after the commit is merged into master, anyway
- The release workflow implemented in this PR won't do the release, it will stop in the last steps with relevant logging information
  - I'll open another PR later for "unlocking" the real release step if everything looks right.

It also changes how commits are tagged:
- uses lightweight tags without messages in the workflow (because short log is already in the release commit body) 
- fix orphan commit on manual script in the least intrusive way I could think of


CHANGELOG.md
```diff
@@ -18,8 +18,6 @@ Changelog
 - Fix link to configuration page (#1020) ([df3b437](https://github.com/dynaconf/dynaconf/commit/df3b43718858786d0a2feb2d7b60c9515b01a15d) by Vladislav Sharapov).
 - Improve Configuration and Validation sections of the docs (#989) ([99741a3](https://github.com/dynaconf/dynaconf/commit/99741a3124b4c788e6a8e9e3761e90e7f9be5cc3) by Sebastian Correa).
 
-<!-- insertion marker -->
```
pedro-psb, line 21: This fixes an error I was having with the new `git-changelog` tool

release.sh
```diff
@@ -146,7 +146,7 @@ git rm -f CHANGELOG.md
 gitchangelog > CHANGELOG.md
 pre-commit run --files CHANGELOG.md || true
 git add -f CHANGELOG.md
-git commit --amend --no-edit
+git commit -m "Update CHANGELOG"
```
pedro-psb, line 149: This fixes the orphan tag, if we need to use the manual script.

.github/workflows/release.yml
```diff
@@ -0,0 +1,111 @@
+name: publish-and-release
+
+on:
```
pedro-psb, line 3: No more on-tag trigger

.github/workflows/main.yml
```diff
@@ -10,22 +10,12 @@ on:
...
       - "docs/**"
       - "mkdocs.yml"
-    tags-ignore:
-      - "**"
-
-  # called by publish workflow to assure pipeline consistency.
-  workflow_call:
-    secrets:
-      CODECOV_TOKEN:
-        required: true
```
pedro-psb, line 19: This was being used to trigger the whole test before running the release workflow.
Reverts what I did [here](https://github.com/dynaconf/dynaconf/commit/c260f52ce4ee194f47753ac8b47e6b75dc16812d)

.github/workflows/doc-ignore.yml
pedro-psb, line 1: This also revert the approach I was having before [here](https://github.com/dynaconf/dynaconf/commit/c260f52ce4ee194f47753ac8b47e6b75dc16812d).

It should allow the test to run after the merge of a tagged commit, as it was before.

.github/workflows/release.yml
```diff
@@ -0,0 +1,111 @@
...
+          name: dist-folder
+          path: dist
+
+      - name: Test publish
+        run: |
+          tree dist
+
+      # - name: Publish package distributions to PyPI
+      #   uses: pypa/gh-action-pypi-publish@release/v1
+
```
pedro-psb, line 87: feature flag, rs

rochacbruno (approved):

PR #1057, opened 2024-02-23
Add python API reference with mkdocstrings

### LINKED ISSUE #997: Add Python API documentation / Module reference to the documentation/website
**Is your feature request related to a problem? Please describe.**
As explained in #996

dynaconf 2.2.3 has a [Module reference](https://dynaconf.readthedocs.io/en/docs_223/reference/dynaconf.html#) where I can read see all the methods of dynaconf.LazySettings together with a description.

I don't see anything similar in https://dynaconf.com/ , @rochacbruno, confirmed that 
> Hi, since we migrated to mkdocs there is **no API reference documented**, there is a mkdocs plugin that can generate that we just need to setup it.

**Describe the solution you'd like**
@rochacbruno  and @pedro-psb  mentioned that there is a mkdocs plugin called [mkdocstring](https://mkdocstrings.github.io/)
that can be used to generate that kind of api documentation . 

So I guess this mkdocs plugin should be configured/setup/included in the pipeline that generates the website

mitches-got-glitches:
--- This PR is in DRAFT ---

Introduces automatic generation of the API reference for `dyanconf` using `mkdocstrings` and [this recipe](https://mkdocstrings.github.io/recipes/#automatic-code-reference-pages).

Adjustments were needed to the template script to get the paths working for a project that was at the root level, rather than inside a `src/` folder. Hopefully these adjustments are explained in the comments.

Building/serving the docs takes quite a bit longer with this automated API reference, around ~80 seconds on my machine. To try for yourselves, check out this branch and explore the served docs. Here's a screenshot of the results.
image

There is definitely a discussion to be had on what is included/excluded as part of the public API, and some adjustments can be made to the script to satisfy more fine-grained requirements. There may be a way to feed in what should/shouldn't be included via some sort of config file. Open to thoughts on this.

Resolves #997

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -19,6 +19,7 @@ nav:
   - Django: django.md
   - Advanced usage: advanced.md
   - FAQ: faq.md
+  - API Reference: reference/
 theme:
   features:
     - search.suggest
@@ -69,6 +70,13 @@ plugins:
   - search:
       lang: en
       prebuild_index: true
+  - gen-files:
+      scripts:
+      - scripts/gen_ref_pages.py
+  - literate-nav:
+      nav_file: SUMMARY.md
+  - section-index
+  - mkdocstrings
 extra:
   disqus: dynaconf
   social:
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,4 +10,8 @@ mkdocs
 pymdown-extensions
 mkdocs-material
 mkdocs-git-revision-date-plugin
+mkdocs-gen-files
+mkdocs-literate-nav
+mkdocs-section-index
+mkdocstrings[python]
 Jinja2>3.0.0
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -40,6 +40,10 @@ mkdocs>=1.1.2
 mkdocs-material>=5.3.2
 mkdocs-material-extensions>=1.0
 mkdocs-git-revision-date-plugin
+mkdocstrings[python]
+mkdocs-gen-files
+mkdocs-literate-nav
+mkdocs-section-index
 pymdown-extensions
 Jinja2>3.0.0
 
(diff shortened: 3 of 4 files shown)


mitches-got-glitches:
I've noticed the docs haven't built on the deploy preview. I'm going to debug in WSL since I was developing on Windows, not Unix. I have a feeling this might be impacting the behaviour of the paths.

pedro-psb:
Hey @mitches-got-glitches, I'm glad you are willing to take this!

I don't think having the code API auto-generated is so important. It looks like it's not so straightforward and we could get it done more quickly if we just declare the modules manually, which shouldn't be too hard as it may seem. Also, we don't add/remove/rename modules too often (mkdocstrings can read everything within a module automatically).

One thing that would be nice to have was to automatically hide anything starting with `_`, if that's possible, because that's the canonical way to say its not public.

For the selection of what goes in there, you could start doing what you think should be public, and we can give some feedback on top of this.

But if you really want to get the automatic generation working feel free, its only a suggestion.

---

> What name would you want to give the top level of the Nav? "API Reference" or "API"?
> Would you want it at the bottom of the existing Nav headers?

Recently I've been a strong addopter of [Diataxis](https://diataxis.fr/). I didn't bring those ideas to dynaconf yet, but this could be a start. So my suggestion is:

```bash
(...) # on the bottom seems fine
Reference/
    API/
    # in the future, we could move dynaconf config, validator listing, chagenlog and similar in here
```

pedro-psb:
> I moved my development to WSL (Linux) and made this change https://github.com/dynaconf/dynaconf/commit/35948c93b87843cf42ea4c0cf2291099eacbf106 - when I serve the docs locally everything builds fine but it still results in a 404 error on the Netlify preview.

My guess is that its not strictly "netlify related", but rather "production build" related, so  you could possibly test it locally by building and running with `python -m http.server -d site/`. Might be easier to debug.

> I might make a separate branch to do it manually rather than reverting, I'll reference it here.

I've just seen it, sounds good.

> I want to try and keep this change atomic, but I am noticing a few warnings on the mkdocs build which could potentially do with addressing so that the output is a little less noisy when you serve.

We are not too strict about keeping those separate, but it's a good practice.

pedro-psb:
I'll close this as suggested by the author in #1058

mitches-got-glitches:
Just for reference, I think this commit c76ca6a  solves the build issue that was occurring with this method.

PR #1058, opened 2024-02-25
docs: Add dynaconf API to docs with mkdocstrings

mitches-got-glitches:
Introduces automatic generation of the API reference for `dyanconf` using `mkdocstrings`.

By simply including a markdown file with:
```md
::: dynaconf
```
... and adding this to the navigation, it autogenerated the API from everything listed in the API in the `__init__.py`:
https://github.com/dynaconf/dynaconf/blob/0390393c27a7ef27104bbda2426b3382dcc7fb9f/dynaconf/__init__.py#L30-L42

This is a lot simpler than what was generated in the attempted PR: #1057 which was following a recipe to build out a reference for all modules within the package. Although it was good learning, I suggest that this one is closed.

From the generated docs, there are probably some follow up changes that can be made to the docstrings to improve the appearance in the API reference. Open to guidance on this.

Resolves: #997


mitches-got-glitches:
These are [the features](https://mkdocstrings.github.io/python/#features) of the Python handler for `mkdocstrings`. Also the look and feel can be tailored in lots of ways with [these options](https://mkdocstrings.github.io/python/usage/#options-summary). Let me know if there's anything worth exploring.

pedro-psb (changes requested):
I think this is a good start.

**styling**
I would suggest these options (which I've seen recently and I think looks good):

```
show_symbol_type_heading: true
show_symbol_type_toc: true
```

And this extra style: https://mkdocstrings.github.io/python/usage/customization/#material
Feel free to play around with styling/display options.

**docstrings**
There is another problem which is that our docstrings are not very consistent. We can start definig [Google Style](https://mkdocstrings.github.io/python/usage/configuration/docstrings/#docstring_style) here, but I'll open another issue about this standardization.

pedro-psb:
> I think it's worth it to get the type hinting consistent and all docstrings following Google style for the current top level API

Yeah, that would be awesome.

> Do you want the dynaconf mod listed at the top of the TOC? This can be switched off if desired.

I would omit that as it does not point to any content.

> Is the full path desired or is that unnecessary clutter?

I'm not 100% sure about this, but I slightly prefer the full path because it's more explicit. We can revisit it later if we feel like its cluttering the view.

pedro-psb:
Also, for the `css` style to work you have to include it using `mkdocs.yml`: https://squidfunk.github.io/mkdocs-material/customization/#additional-css

mitches-got-glitches:
Thanks again for the feedback, I've made a couple more changes. 

> > I think it's worth it to get the type hinting consistent and all docstrings following Google style for the current top level API
> 
> Yeah, that would be awesome.

Do you want me to try and get this in shape for the top level API on this ticket so that it looks good when merged in? It can form part of the discussion on #1060.

pedro-psb (approved):
@mitches-got-glitches 
edit:
> Do you want me to try and get this in shape for the top level API on this ticket so that it looks good when merged in?

There is no need. It will only get published when we release, and even so, I wouldn't bother having it published.
Can I merge it?

docs/api.md
```diff
@@ -0,0 +1,4 @@
+# dyanconf API
```
ecerulm, line 1: It should be `dynaconf` not `dyanconf`

ecerulm:
@pedro-psb , there is a typo `dyanconf API` instead of the `dynaconf API`. 

Also when it's this supposed to appear on https://dynaconf.com ? I don't see it yet.

mitches-got-glitches:
> @pedro-psb , there is a typo `dyanconf API` instead of the `dynaconf API`. 
> 
> Also when it's this supposed to appear on https://dynaconf.com ? I don't see it yet. 

Well spotted, I will fix it on #1062.

I think a new release version needs to be published before dynaconf.com is updated.

 [typos](https://github.com/crate-ci/typos) may be worth looking at in the near future to prevent spelling errors creeping in.

ecerulm:
> I think a new release version needs to be published before dynaconf.com is updated.
Last release was on  2023-10-30 according to  https://github.com/dynaconf/dynaconf/blob/master/CHANGELOG.md

Almost 4 months ago, do we know when is the next planned release?

pedro-psb:
@ecerulm Thanks for catching that. There is not a fixed date but we are planning to do it soon, maybe in the next week or so.

mitches-got-glitches:
I've just started a blog and I wrote up on my learnings completing this PR. Here it is for future reference!
https://mitches-got-glitches.github.io/developer_blog/

pedro-psb:
@mitches-got-glitches I've just read it, that's a great storytelling flow! I'm glad this was a valuable learning experience for you.
I feel like knowing this kind of skill (doc tooling) makes someone able to contribute to so many projects that require that kind of improvement.
And it is good to hear that the "good first issue" works :sweat_smile:

PR #826, opened 2022-10-30
Build translated docs to fix #797

winmorre:
To test this PR 

run `python ./scripts/docs.py build-all`  This will build the docs site

run `python ./scripts/docs.py serve`  This will serve the doc site

Issue #797, opened 2022-09-02
[RFC] Build translated docs


rochacbruno:
Now inside `docs/` there are subfolders for each translated language,

`docs/pt-br/docs`

The goal now is to prepare MKDOCS to serve in English docs on `/` but serve portuguese on `/pt-br`

We need to customize the build step currently called `make docs` to make that happen and produce the expected output on the `site` folder.

Also it would be good to add translations to the menu

NOTE: See how FastAPI makes it

winmorre:
Can I pick this up?

rochacbruno:
sure @winmorre

rochacbruno:
I wanna solve this, the PR looks good, I need to carefully review it.

winmorre:
sure, let me know if  there is something I need to do

PR #1061, opened 2024-02-26
docs: Fix mkdocs warnings for cleaner build output

mitches-got-glitches:
Working on a recent PR, the output from `mkdocs build` and `mkdocs serve` was very noisy because of warnings. This PR fixes them all to make the build clean. There were many like the first screenshot.
image
image
image

### Result
image

Much cleaner, only a few warnings around the Portuguese language docs which I wasn't sure how to deal with.

### Other comments

- My IDE automatically deleted some trailing whitespace from the markdown files on save
- I'm not sure if the previous version of the references was actually useful within the Portuguese docs? Can you advise how these are currently in use?

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -79,7 +79,7 @@ settings = Dynaconf(post_hooks=hook_function)
 ```
 
 You can also set the merging individually for each settings variable as seen on
-[merging](/merging/) documentation.
+[merging](merging.md) documentation.
 
 ## Inspecting History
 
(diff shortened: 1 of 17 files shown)


pedro-psb (approved):
Great! Thanks for the contribs.

I was unsure about `prebuild_index`, but it looks like its [not supported](https://github.com/squidfunk/mkdocs-material/discussions/4764#discussioncomment-4408399) for some time in mkdocs-material.

pedro-psb:
> I'm not sure if the previous version of the references was actually useful within the Portuguese docs? Can you advise how these are currently in use?

These files are not currently being used. They were suppose to prepare for a translated build, which was discussed in #797  and was being developed in #826.

PR #1063, opened 2024-03-01
Introduce typos pre-commit and fix all current typos

mitches-got-glitches:
Hey guys, this is optional but I've found this tool to be quite good. It's built in rust so it's fast and I've found it to be pretty accurate. However, in the pre-commit I've put it to warn, rather than write the changes automatically so that the developer has the choice to configure if needed. Let me know if there are other sections that you'd want excluded.

https://github.com/crate-ci/typos/tree/master

### Usage

```typos```

```typos -w    # Writes the changes```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ on:
   workflow_dispatch:
 
 # When a Push is made on an existing branch/PR
-# it cancells any pre running job and
+# it cancels any pre running job and
 # let only the latest executing
 concurrency:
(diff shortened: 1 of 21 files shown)


pedro-psb (approved):
Oh my god, so many typos. It really works!
+1 for just warning, by default. After some experimentation, we might change that, if we feel its safe.

Nice work.

rochacbruno (approved):

mitches-got-glitches:
Yeah I was very impressed to come across one that threw so few false positives, particularly when scanning code!

PR #1070, opened 2024-03-05
fix incorrect combination of TOML table and inline table

adevore:
The `settings.toml` example for initializing Dynaconf includes an inline table and a separate section that looks like it should nest. However, TOML's inline tables must be entirely self contained, with no members in a separate section.


pedro-psb (approved):
Yeah, it really raises an error.
Thanks

Issue #1060, opened 2024-02-25
Standard docstrings style for the codebase


pedro-psb:
## Problem

Some parts of the codebase use Sphinx-style docstrings and others use Google's, and there is not a clear definition about this. Also, with the addition of auto-generated docs (#1058) , the docs render doesn't come out very pretty.

See [here](https://bwanamarko.alwaysdata.net/napoleon/format_exception.html) examples of these (and other) different styles.

## Proposal

- Define Google Style as the default style
- Change all docstrings to the new format
- Use some linter/formatter to enforce this with out pre-commit setup. (e.g. https://pypi.org/project/docformatter/)

## Alternatives

We could use another style as well (if there is a strong opinion towards that). I would only argue that using a linter/auto-formatter should help get the codebase sharp.

mitches-got-glitches:
Hey @pedro-psb , I wanted to get your thoughts on using [ruff](https://github.com/astral-sh/ruff) for docstring standardisation instead of `docformatter`? `ruff` could be brought in for this use case, but eventually it could be swapped in as a replacement for `flake8`, `black` and import sorter too.

rochacbruno:
+1 for google style

pedro-psb:
I didn't know ruff did that, I'm ok with giving it a try.
Wdyt @rochacbruno ?

pedro-psb:
Yes, we could use that.
It requires some research on the tool. The requirements are:
- It must be a linter: it must at least check that the docstring complies with Google Style.
- It may be a formatter. If it cannot autoformat, that is ok
- It should run on pre-commit: if it absolutely can't, we can look for another way of integrating it

If ruff can do that, we should use it. I'm not sure it can, its worth looking.
Feel free to ask for help if you need it.

skritosss:
Ran ruff with convention = "google" over your code to get a sense of the scope. Didn't count the vendor directory, so 40 files.
The full D set gives 566 errors. But around 200 of those are D100–D107, i.e. no docstring at all.
With D1xx off and only the style rules left, it's 363 errors, and ruff fixes 294 of them itself. That leaves roughly 70 spots to touch by hand, mostly a missing period at the end of the first line and a missing blank line after the summary.
Ruff covers both of your requirements: it checks Google Style and runs through pre-commit, so docformatter isn't needed.
One question before I start: do D1xx stay off, or should a missing docstring count as an error too? That decides whether this is one PR or several.

Happy to take it on if the direction is right.

PR #1074, opened 2024-03-06
ci: Replace lint and formatting tools with ruff

mitches-got-glitches:
### Context

I thought I would have a go at adding ruff to the codebase in anticipation of using it for #1060. My initial thought was to suggest [`pydocstyle`](https://github.com/PyCQA/pydocstyle) for that ticket, but this project has been deprecated in favour of `ruff`. The speed benefits of using `ruff` are pretty significant:
image

Throughout doing this ticket, I have figured out that we could add `ruff` and tell it only to check the `pydocstyle` codes - so that is an alternative if this change is a bit much.

### Key changes

- `ruff` replaces `black`, `flake8`, `reorder-python-imports` and some other pre-commit hooks in [`.pre-commit-config.yaml`](https://github.com/dynaconf/dynaconf/compare/master...mitches-got-glitches:dynaconf:ci/add-ruff?expand=1#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R7-R21)
- `ruff` added to [`requirements_dev.txt`](https://github.com/dynaconf/dynaconf/compare/master...mitches-got-glitches:dynaconf:ci/add-ruff?expand=1#diff-c9c796259f3852b51b531b79cbf07820145088d4f108fb006ab44b15f0b15934R22-R26). `flake8`, `pep8-naming`, `flake8-debugger`, `flake8-print`, `flake8-todo` are removed.
- Config added to top level [`ruff.toml`](https://github.com/dynaconf/dynaconf/compare/master...mitches-got-glitches:dynaconf:ci/add-ruff?expand=1#diff-26f436cb29eecc7bb4f9066747b43df8da81ac808955ce95c109f5d7aa778989R1-R32) to mirror previous configuration as closely as possible.
- Lower level configs added under [`tests/ruff.toml`](https://github.com/dynaconf/dynaconf/compare/master...mitches-got-glitches:dynaconf:ci/add-ruff?expand=1#diff-adc11f0dd9c9794987b40597334f9440f34ca3e95250d0b74cde7187a1467f12R1-R8) and [`tests_functional/ruff.toml`](https://github.com/dynaconf/dynaconf/compare/master...mitches-got-glitches:dynaconf:ci/add-ruff?expand=1#diff-e42efa3d7189568b729f98e6ce58e21dce0161a18bdebe1e61982dd6ee0c0370R1-R9) to prevent the need for unnecessary changes to these packages. These can be reviewed at a later time.
- [`pep8` section removed from Makefile](https://github.com/dynaconf/dynaconf/compare/master...mitches-got-glitches:dynaconf:ci/add-ruff?expand=1#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L66-L73) - covered instead by the `pre-commit` section

### Notes

- [Rules W503 and W504](https://github.com/dynaconf/dynaconf/compare/master...mitches-got-glitches:dynaconf:ci/add-ruff?expand=1#diff-26f436cb29eecc7bb4f9066747b43df8da81ac808955ce95c109f5d7aa778989R21-R25) are not implemented in Ruff - see this issue: https://github.com/astral-sh/ruff/issues/4125
- Many of the changes made can be explained by looking at `ruff`'s [Known Deviations from Black](https://docs.astral.sh/ruff/formatter/black/)
  - [This change](https://github.com/dynaconf/dynaconf/compare/master...mitches-got-glitches:dynaconf:ci/add-ruff?expand=1#diff-d4d1e188f1582e0dcc9d59f05c3688706aa1f5dafbd605d9ce0f57bfab07aa4dR394-R395) happens because ruff [looks at Unicode width](https://docs.astral.sh/ruff/formatter/black/#line-width-vs-line-length)
  - Adding a [line of whitespace after module docstrings](https://github.com/dynaconf/dynaconf/compare/master...mitches-got-glitches:dynaconf:ci/add-ruff?expand=1#diff-b6ca14946f7a88707df6e31bcff95821f2a8f54919bb48c9506422c1b7111683R13-R17) seems to be something `ruff` wants to do (can't find the explanation), and this amounts to quite a lot of the changes
  - [This change](https://github.com/dynaconf/dynaconf/compare/master...mitches-got-glitches:dynaconf:ci/add-ruff?expand=1#diff-b6ca14946f7a88707df6e31bcff95821f2a8f54919bb48c9506422c1b7111683R13-R17) is because [pragma comments are ignored when computing line width](https://docs.astral.sh/ruff/formatter/black/#pragma-comments-are-ignored-when-computing-line-width)
  - [This is one change](https://github.com/dynaconf/dynaconf/compare/master...mitches-got-glitches:dynaconf:ci/add-ruff?expand=1#diff-3228fac294f31aa71060654831c3e0d21cef6076de683835bce5ac31334e321eR302-R305) that looks a little buggy to me. The `# type: ignore` has moved to the second broken line when I'm guessing it has been added for the `_ensure_serializable(v)`. Didn't seem to be picked up by `mypy` though so it still passes. Should this be moved up a line in this case? I'm not sure [this section on pragma comments](https://docs.astral.sh/ruff/formatter/black/#pragma-comments-are-ignored-when-computing-line-width) explains the behaviour we're seeing here enough.
- `ruff` seems to be [one-lining these](https://github.com/dynaconf/dynaconf/compare/master...mitches-got-glitches:dynaconf:ci/add-ruff?expand=1#diff-2bf0e4b7600a8b9bacd98749033f5ad209b918713653b22625bdd76fb9f36fd5R27) and this has been [raised as an unintentional deviation](https://github.com/astral-sh/ruff/issues/10186). Maybe something that the project can live with for now?
- It looks like `black` was never in `requirements_dev.txt`?

### Extras
- The ignore rules in the `ruff.toml` can potentially be tuned as [`per-file-ignores`](https://docs.astral.sh/ruff/settings/#lint_per-file-ignores)
- I can offer some extra proof that some of the things that have been removed still work with `ruff` - just let me know what you want to see.


dynaconf/utils/inspect.py
```diff
@@ -298,7 +300,8 @@ def _ensure_serializable(data: BoxList | DynaBox) -> dict | list:
         return [_ensure_serializable(v) for v in data]
     elif isinstance(data, (DynaBox, dict)):
         return {
-            k: _ensure_serializable(v) for k, v in data.items()  # type: ignore
+            k: _ensure_serializable(v)
+            for k, v in data.items()  # type: ignore
```
pedro-psb, line 304: We have a loose mypy config, it doesn't comply even if we remove the comment.
This is fine.

dynaconf/hooking.py
```diff
@@ -25,8 +24,7 @@
 ]
 
 
-class Empty:
-    ...
+class Empty: ...
```
pedro-psb, line 27: Thats ok.
We could probably have a simple docstring for this (later) which should revert this.
Something like "Sentinel that represents an empty value".

.pre-commit-config.yaml
pedro-psb, line 1: lgtm

requirements_dev.txt
pedro-psb, line 1: yeah, we didn't have black :open_mouth:

pedro-psb (changes requested):
That looks good. I can see it catches TODO, breakpoints and prints, and seems to do a good job at sorting the imports. Also, I don't mind the known style deviations.

Can you hard-comment the reason for `tests/ruff.toml` and 'test_functional/ruff.toml` ? I think it can be helpful.

pedro-psb:
We should wait @rochacbruno review as well, as these are pretty significant changes.

And out of curiosity, why would we want per-file-ignores? I'm not saying we don't, I just don't know.

mitches-got-glitches:
> We should wait @rochacbruno review as well, as these are pretty significant changes.
> 
> And out of curiosity, why would we want per-file-ignores? I'm not saying we don't, I just don't know.

Just as an example, if I enable a check on E402: Module level import not at the top of the file, we end up with these issues reported back:
image

Now we might not care too much about the rule contraventions in `tests_functional/` but maybe we do care about it in `dyanconf/cli.py`. So we could have something like this in the `ruff.toml`:

```toml
[lint.per-file-ignores]
"tests_functional/*" = ["E402"]
```

And we end up with this:
image

The checks that we currently ignore may be worth applying over the `dynaconf/` folder at least. What do you think?

mitches-got-glitches:
> Can you hard-comment the reason for `tests/ruff.toml` and 'test_functional/ruff.toml` ? I think it can be helpful.

I've added some new commits which actually remove these in favour of per-file-ignores. Let me know what you think of the latest commits where I've implemented per-file-ignores. Tried to do everything on separate commits so things can be rolled back if needed.

pedro-psb (approved):
Thanks for clarifying per-ignore-files.
Its a nice feature and I prefer it over the extra config files.
This looks good for me! :tada:

rochacbruno (approved):

PR #1079, opened 2024-03-19
chore(ci): Add workflow to publish docs to netlify

pedro-psb:
What this does:
- Add workflow to publish docs to netlify through [Netlify GitHub action](https://github.com/marketplace/actions/netlify-actions#usage) (instead of using the Netlify App)
  - The workflow is triggered on semver tags and through workflow dispatch.
  - The manual dispatch provides an option to checkout to another refspec before building the docs (in the case a building an on-tag build failed for some reason).
- Comment out the `netlify.toml` build definitions, so the App won't interfere with Github actions.

### DIFF
--- a/.github/workflows/netlify.yml
+++ b/.github/workflows/netlify.yml
@@ -1,0 +1,51 @@
+name: Build and Deploy to Netlify
+
+on:
+  push:
+    branches: [master]
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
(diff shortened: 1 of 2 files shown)


rochacbruno (approved):
Do we need to define NETLIFY_AUTH_TOKEN and NETLIFY_SIDE_ID?

pedro-psb:
Lets see if that works :eyes:

PR #1080, opened 2024-03-22
Fix #1055 allow multiple validators with cast for the same field

### LINKED ISSUE #1055: Multiple cast validators get discarded
Hi there, thanks for this great project,

**Describe the bug**

After defining a dynaconf `cast` validator, attempting to define subsequent dynaconf `cast` validators on the same variable get discarded while they should also be taken into account

**To Reproduce**

1. Having the following a.toml file:

**a.toml**
```toml
var = "1"
```

2. And the following sample code:

**a.py**
```python
from dynaconf import Dynaconf, Validator

settings = Dynaconf(settings_file="a.toml")


def a(v):
    print("called a")
    return v


def b(v):
    print("called b")
    return v


validators = [
    Validator("var", cast=a),
    Validator("var", cast=b),
]

settings.validators.register(*validators)
settings.validators.validate_all()
```

3. Executing `a.py` under a virtualenv with `dynaconf`  `3.2.4` installed:

```bash
$  python3 a.py
called a
```

I would expect to see `called a` and then an additional line `called b` after that.

**Additional context**

This bug is because of `Validator("var", cast=a)` and `Validator("var", cast=b)` comparing equal with the current definition of [`__eq__`](https://github.com/dynaconf/dynaconf/blob/0390393c27a7ef27104bbda2426b3382dcc7fb9f/dynaconf/validator.py#L155-L169):

```python
    def __eq__(self, other: object) -> bool:
        if self is other:
            return True


        if type(self).__name__ != type(other).__name__:
            return False


        identical_attrs = (
            getattr(self, attr) == getattr(other, attr)
            for attr in EQUALITY_ATTRS
        )
        if all(identical_attrs):
            return True


        return False
```

and with 

```python

EQUALITY_ATTRS = (
    "names",
    "must_exist",
    "when",
    "condition",
    "operations",
    "envs",
)
```

Since `cast` is not in the `EQUALITY_ATTRS`, validators on the same name with just a cast defined are considered equal and the second validator are skipped altogether (see [this part of the code](https://github.com/dynaconf/dynaconf/blob/0390393c27a7ef27104bbda2426b3382dcc7fb9f/dynaconf/validator.py#L463))

I think a one-line fix for this would be to add the `cast` attribute to the EQUALITY_ATTRS.

rochacbruno:
```python
from dynaconf import Dynaconf

settings = Dynaconf(var="dynaconf")

def a(v):
    print("called a, replace conf -> settings")
    return v.replace("conf", "settings")

def b(v):
    print("called b, transform to upper DYNASETTINGS")
    return v.upper()

validators = [
    Validator("var", cast=a),
    Validator("var", cast=b),
]

settings.validators.register(*validators)
settings.validators.validate_all()

assert settings.var == "DYNASETTINGS", settings.var
```

Fix #1055

### DIFF
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -350,6 +350,10 @@ assert settings.colors == '["red", "green", "blue"]'
 assert type(settings.colors) == str
 ```
 
+> **NOTE**: When multiple validators for the same field specifies a `cast`, Dynaconf
+> will execute both in order, so the end result in the settings object is the result
+> of the cumulative pipeline.
+
 ### Custom conditional expressions
 
 The `condition` argument expects a `Callable` that receives the setting's value as its only argument and returns a `bool`.
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -350,9 +350,8 @@ assert settings.colors == '["red", "green", "blue"]'
 assert type(settings.colors) == str
 ```
 
-> **NOTE**: When multiple validators for the same field specifies a `cast`, Dynaconf
-> will execute both in order, so the end result in the settings object is the result
-> of the cumulative pipeline.
+!!! info
+    When multiple validators for the same field specifies a `cast`, Dynaconf will execute both in order, so the end result in the settings object is the result of the cumulative pipeline.
 
 ### Custom conditional expressions
 
--- a/dynaconf/validator.py
+++ b/dynaconf/validator.py
@@ -16,6 +16,10 @@ if TYPE_CHECKING:
     from dynaconf.base import LazySettings  # noqa: F401
     from dynaconf.base import Settings
 
+
+DEFAULT_CAST = lambda value: value  # noqa
+
+
 EQUALITY_ATTRS = (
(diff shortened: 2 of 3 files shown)


docs/validation.md
```diff
@@ -350,6 +350,10 @@ assert settings.colors == '["red", "green", "blue"]'
 assert type(settings.colors) == str
 ```
 
+> **NOTE**: When multiple validators for the same field specifies a `cast`, Dynaconf
+> will execute both in order, so the end result in the settings object is the result
+> of the cumulative pipeline.
+
```
pedro-psb, line 356: We should use the [admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#usage) for notes:
```
!!! note

    When multiple (...)
```

pedro-psb (changes requested):
Otherwise LGTM

pedro-psb (approved):

PR #1091, opened 2024-04-19
Fix #1088 fresh vars definition case insensitive

### LINKED ISSUE #1088: [bug] fresh_vars items are assignable
**Describe the bug**
Assigning to a setting that exists in `fresh_value` prevents that setting from being fresh.

**To Reproduce**
Minimum code to reproduce/demonstrate:

```
import os
from dynaconf import Dynaconf

os.environ["DYNACONF_FRESH_VALUE"] = "stay fresh"
settings = Dynaconf(fresh_vars=["fresh_value"])
print(settings.fresh_value) # Outputs: "stay fresh"
settings.fresh_value = "not so fresh"
print(settings.fresh_value) # Outputs: "not so fresh" instead of "stay fresh" as expected.
```

**Expected behavior**
From the documentation, I'd expect that any attempt to assign to a fresh variable would be passively ignored, since the docs state that the value is read directly from source on every access. Apologies if this is WAI. If it is, it may be worth noting on the documentation for the `fresh_vars` parameter that once assigned to the parameter will no longer be fresh.

 - OS: Windows/10 & Linux/Ubuntu22.04.4 (WSL)
 - Dynaconf Version: 3.2.5
 - Frameworks in use: None.

**Additional context**

rochacbruno:
Fix #1088


tests_functional/issues/settings.toml
pedro-psb, line 1: This one was left over

pedro-psb (changes requested):

pedro-psb (approved):

Issue #1108, opened 2024-06-19
[RFC] Suport sub types


rochacbruno:
#1107 introduces tech preview of Typed Settings

Now the next feature to be added is SubTypes

right now every attribute on typed.Dynaconf must have a type annotation or a VAlidator, but it doesnt allow yet to have a custom type.

```
 class Database:       
    host: str       
    port: Annotated[int, Validator(gt=999)] 

 class Settings(Dynaconf):      
    database: Database
```

  For this to work, we might need to have the subschema to also contribute to Validators and its value casted, because on the raw settings there will be `{"database": {...}}` and not the instance of Database

pedro-psb:
Closed by #1109

PR #1109, opened 2024-06-19
fix: Support Nested Subtypes and Enclosed types

rochacbruno:
related #1108

> WARNING: This code is not perfect yet, we can merge because we are in tech preview for thisw feature, however the implementation is bad and needs refactoring.

```python
    class Database(Nested):
        host: str = "server.com"
        port: Annotated[int, Validator(gt=999)]
        engine: str = "postgres"
        versions: list[int] = [1, 2, 4]
        extra: Optional[dict]

    class Settings(Dynaconf):
        dynaconf_options = Options(
            envvar_prefix="MYTYPEDAPP",
        )
        database: Database
        batata: Annotated[int, Validator(gt=999)]

    os.environ["MYTYPEDAPP_BATATA"] = "1000"
    os.environ["MYTYPEDAPP_DATABASE__PORT"] = "5000"
    settings = Settings()

    assert settings.database.port == 5000
    assert settings.database.host == "server.com"
    assert settings.database.engine == "postgres"

```


dynaconf/typed.py
```diff
@@ -92,21 +83,197 @@ def __new__(cls, *args, **kwargs):
...
+    Nested,
+    int,
+    float,
+    str,
+    bool,
+    type(None),
+    list,
+    tuple,
+    dict,
+)
```
pedro-psb, line 125: Its nice that we declare what is explicitly supported.

dynaconf/typed.py
```diff
@@ -92,21 +83,197 @@ def __new__(cls, *args, **kwargs):
...
+            )
+        # It is a field: Annotated[type, Validator(...)]
+        elif _annotated_validators := get_annotated_metadata(annotation):
+            if isinstance(_type, type) and issubclass(_type, Nested):
+                raise DynaconfSchemaError(
+                    "Nested type cannot be Annotated, "
+                    f"set the validators on the {_type.__name__!r} fields "
+                    f"this error is caused by "
+                    f"`{full_name}: Annotated[{_type.__name__}, *]`"
+                )
```
pedro-psb, line 183: Not allowing `Annotated[Nested, ...]` is a good catch too.
That could get confusing quickly

dynaconf/typed.py
```diff
@@ -92,21 +83,197 @@ def __new__(cls, *args, **kwargs):
...
+            _validator = Validator(full_name, is_type_of=annotation)
+            if default_value is empty and not is_optional(annotation):
+                _validator.must_exist = True  # required
+            validators.append(_validator)
+
+    return defaults, validators
+
+
+def validate_enclosed_list(validators, prefix, _type):
+    """takes list[_type] and test against validators"""
```
pedro-psb, line 236: I would name that `enclosed_list_validator_factory`, `create_enclosed_validator_fn`, `create_validator_conditon` or something like that.
I was a bit confused thinking this would really validate something at this point.

dynaconf/typed.py
pedro-psb, line 1: That's looking good.
I really liked the comments on the recursion cases, they were very helpful!

tests/test_typed.py
```diff
@@ -64,3 +65,229 @@ def test_options_contains_only_valid_and_set_attributes():
...
+        Optional, but if set, value types are validated
+            Optional[list[T]]
+
+        Required, value types validated and can't be an empty list
+            list[T]
+
+        Required, must be a list, but can be an empty list,
+        If has values, value types are validated
+            list[Optional[T]]
+        """
```
pedro-psb, line 173: `list[Optional[T]]` is a bit ambigous. Reading "optional" as usual conveys the idea that its optional to pass the enclosed value, but thinking about it as `Union` makes us think we can do `["foo", None, None, "bar"]`.

Some alternative ideas I had today:
- provide a custom type `optionalList`, `emptyList`, `optionalEmptyList` (cant think of a good name) and define this specific behavior.
- use a new spec obj on the Annotated arguments, like `key: Annotated[list[str], ListValidator(min_size=0)]`.

pedro-psb (approved):
This is looking good, the error handling is great and its going to be a very cool feature!
I've commented a few nitpicks, but I'm really happy to merge that as it is.

dynaconf/typed.py
rochacbruno: working on it!

dynaconf/typed.py
rochacbruno: The idea for that type is just to be a marker for something that is Nested, like a `Struct` that at the end will be a `dict`, we could name that something that reminds a dict.

- Struct
- Map
- DictValue

pedro-psb: Yeah, maybe Schema doesnt convey that idea very clearly...
Naming things is so hard. I'm happy with what you feel is best.

tests/test_typed.py
rochacbruno: What about `Nullable` ?

pedro-psb (approved):

PR #1120, opened 2024-07-03
misc:Add one functional test for typed

rochacbruno:


tests_functional/typed/typed_with_dotenv_toml_secrets_validator_condition/.env
```diff
@@ -0,0 +1,3 @@
+MYAPP_TITLE="My Awesome App"
+MYAPP_PLUGINS="@merge [{name='tool'}]"
```
pedro-psb, line 2: I didn't know we could do that. That's effectively an append.

pedro-psb (approved):

PR #1125, opened 2024-07-05
Validators aliases

rochacbruno:
Inspired by the module https://github.com/annotated-types/annotated-types added some better named validators.

Also reorganized validation objects under typed/validators.py

```python
from typing import Annotated
from dynaconf.typed import Dynaconf
from dynaconf.typed.validators import Eq, Contains, IsUrl, Regex, Not

class Settings(Dynaconf):
    fruit: Annotated[str, Not(Eq("banana"))]
    static_url: Annotated[str, IsUrl()]
    version: Annotated[str, Regex(r"^(\d+).(\d+).(\d+)$")]
    colors: Annotated[list[str], Contains("red")]
```

all the above are just aliases, which is equivalent to

```python
from typing import Annotated
from dynaconf.typed import Dynaconf, Validator

class Settings(Dynaconf):
    fruit: Annotated[str, Validator(ne="banana"))]
    static_url: Annotated[str, Validator(regex=r"((http|https)://)...")]
    version: Annotated[str, Validator(regex=r"^(\d+).(\d+).(\d+)$")]
    colors: Annotated[list[str], Validator(contains="red")]
```

### DIFF
--- a/dynaconf/typed/validators.py
+++ b/dynaconf/typed/validators.py
@@ -211,7 +211,7 @@ class Not(BaseValidator):
 # Generic Validation Aliases
 
 
-class GenericParametrizedOperation:
+class GenericOperation:
     """Base to create aliased Validator with generic parametrized operation.
 
     This class exists mainly to allow creation of semantic named validators.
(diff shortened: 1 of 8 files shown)


dynaconf/validator_conditions.py
```diff
@@ -89,46 +90,78 @@ def is_type_of(value, other):
...
 
 
-def cont(value, other):
+def contains(value, other) -> bool:
     """Contains"""
     return other in value
 
 
-def len_eq(value, other):
+cont = contains  # backwards compatibility
```
pedro-psb, line 108: :+1:

dynaconf/typed/validators.py
```diff
@@ -0,0 +1,449 @@
...
+    Which is equivalent to::
+
+       IsAFruitILike = Validator(is_in=["banana", "apple", "lemon", "orange"])
+       fruit: Annotated[str, IsAFruitILike]  # not callable
+
+    However simply assigning the validator to a variable doesnt allow for
+    inheritance and doesn't create a callable object, being a callable object
+    is useful to keep the possibility to accept arguments if needed without
+    breaking compatibility.
+    """
```
pedro-psb, line 401: We could have a dedicated Validator alias constructor to normalize this, but I dont think we need to.

If we are adding this new semantic system, lets prioritize it as the recommended and most supported interface for the users, and the Validator becomes a more or less internal interface.

pedro-psb (approved):

PR #1126, opened 2024-07-06
misc(typed):Renamed some validators and parameters

rochacbruno:
- Renamed `gte` to `ge` and `lte` to `le` for better matching with Python's `__ge__` etc...
- Renamed `GenericOperation` -> `GenericValidator`
- Replaced `CustomOperation` with just regular variables, it is ok to not be callable if it doesn't accept arguments `Annotated[T, is_url]` is ok and when there are args `Annotated[T, Eq(x)]`
- Made `GenericValidator` to accept partial arguments


dynaconf/typed/validators.py
```diff
@@ -232,218 +232,203 @@ class NoWayItIs(GenericOperation):
...
+                return (
+                    any(c in val for c in "@#$&!%")
+                    and any(c.isupper() for c in val)
+                    and any(c.isdigit() for c in val)
+                )
+
+            args = [contains_special_chars]
+
+        field: Annotated[str, GoodPassword(10, 20)]
+
```
pedro-psb, line 258: Cool :)

pedro-psb (approved):

Issue #1158, opened 2024-07-08
Instantiating Dynaconf can be slow


sebastian-correa:
#Summary
I'll preface with a _thank you_ for the work you've done in Dynaconf. We've been using it for a while now and it's helped our dev team considerably.
 
We're using Dynaconf to load ~14 YAML files. Due to our package's `__init__.py`, Dynaconf is one of the first things that loads and it's become a bit too slow for us, as startup takes ~2s-4s most of which is Dynaconf:

<img width="544" alt="image" src="https://github.com/dynaconf/dynaconf/assets/43179146/f799831e-4411-4e2a-bc59-f8b0fa863ace">

This is how we declare the `Dynaconf` object:

```python
settings = Dynaconf(
    settings_files=_list_of_14_yamls,
    environments=True,
    env_switcher="PROJECT_STAGE",
    envvar_prefix="PROJECT",
    merge_enabled=True,
    post_hooks=[cast_settings],
)
```
where the `cast_settings` function is a simple function that does things like `settings["some"]["setting"] = SomeEnum[settings["some"]["setting"]]`.

I've been profiling our code and found that the biggest culprits seem to be `Validator`s and `execute_module_hooks`. In our case, these account for ~0.8s-1.5s and ~0.7s of startup time. The casting function takes about ~0.4s as well, but that seems unavoidable in our case (not using casting from `Validator`s because it gave trouble some time ago; I haven't checked in recent versions).

## Validation
Of the validation time, 2 `Validator`s take most of the time (one that checks ~22 settings takes ~0.4s and another with ~11 takes ~0.2s):

```python
slow_validators = [
    Validator(*long_list, cast=float, default=0.0),
    Validator(*long_list_2, must_exist=True, cast=float),
]
```

Does this runtime for each of these validators make sense to you?

## `execute_module_hooks`
The call to `execute_module_hooks` in `Settings.__init__` seems to want to import 1 `dynaconf_hooks.py` file per YAML we tried to load. This results in calls to `py_loader.import_from_filename`, which takes some time.

For this, I propose to add

```python
    if not os.path.exists(filename):
        # Return early if file doesn't exist (can't be imported). Faster than attempting to import.
        return
```

at the very top of `import_from_filename`. I can contribute this with a PR if it's something you'd accept though I could also research this function further to find the actual culprit of the delay. In our case, this solution brought down runtime from ~0.7s to ~0.1ms.

rochacbruno:
[...]

PR #1159, opened 2024-07-09
fix(hooks): Ignore if file doesn't exist

rochacbruno:
Related to #1158


sebastian-correa (approved):
🚀 🛸

Tested by running

```python
from dynaconf import Dynaconf

settings = Dynaconf(settings_files=["test_settings.yaml"])

print(settings["foo"])
```

and everything ran fine.

PR #1160, opened 2024-07-10
docs(hook):Add documentation and example of Access Hooks

### LINKED ISSUE #1095: [docs] Document the new method hooks
Add documentation for the feature added on https://github.com/dynaconf/dynaconf/pull/975

Dynaconf allows changing the wrapped class to allow method hooking.

```python
from dynaconf import Dynaconf

def lookup_for_key_in_cache_or_database(temp_settings, value,  key, default):
    if key.upper() not in [list of keys I am interested]:
        return value.value
    return new_value
    # NOTE: if value is a mergeable data structure, then the temp_settings might be used.
    # temp_settings.set(key, new_value)
    # return temp_settings.get(key, value.value) 

settings["_registered_hooks"][Action.AFTER_GET] = [lookup_for_key_in_cache_or_database]
```

When `settings["_registered_hooks"]` is set as a dictionary, keyed by `Action.AFTER_GET|BEFORE_GET|etc`

at any access to `settings.*`  dynaconf after performing its usual get process, will call the hook functions, chained in sequence, passing down the value from one hook to another, this way the latest hook in the list is the final value.

`alter_settings_from_request(lookup_for_key_in_cache_or_database(do_another_thing(value, **kwargs)))`

Each hook function may do a conditional check (that is why we do if key in list_of_interesting_keys) and if the key is not interesting for the hook it may return value to be passed to the next hook unaltered.

If key is the key that the hook wants, then it can alter it and return the altered value (which will be passed to the next hook in the pipeline) or can return a EagerValue(thing)  that will avoid repassing to the next and just return as the final value.

In our case, I think we want to maintain a list of enabled hooks, and if there are none registered, we can just bypass its configuration.

This all could be registered on app/settings.py directly, or in special cases using dynaconf_hooks.py

rochacbruno:
fix #1095

feature added on 3.2.2 https://github.com/dynaconf/dynaconf/releases/tag/3.2.2

## Access Hooks

Dynaconf can be configured to retrieve a special variable from an external
storage such as GCP or AWS secrets or a Vault server or even a secret located
on the filesystem or encrypted.

This works by registering a function hook that will be called for every variable
access on Dynaconf.

This feature is useful to load settings from BitWarden and other secret managers.

NOTES:
   - You must define the logic of each hook, dynaconf just calls it
   - You must take care of caching and caching invalidation
   - This may introduce overhead dependending on how the hook function is
      implemented

```python
# So you have a function that retrieves data from external source

CACHE = {}  # could be Redis

def get_data_from_somewhere(settings, result, key, *args, **kwargs):
    """Your function that goes to external service like gcp, aws, vault, etc
    NOTE: settings in this context is copy of settings, changes made are
    valid only during the scope of this function.
    """
    if key in CACHE:
        return CACHE[key]

    # Assuming result.value is '@special:/foo/bar:token'
    if isinstance(result.value, str) and result.value.startswith("@special:"):
        _, path, identifier = result.value.split(":")
        # value = get_value_from_special_place(path, identifier)
        value = CACHE[key] = "lets believe it was retrieved from special place"
        return value

    return result

# ---

from dynaconf import Dynaconf
from dynaconf.hooking import HookableSettings, Hook, Action

# Create a Dynaconf instance
settings = Dynaconf(
    # set the Hookable wrapper class
    _wrapper_class=HookableSettings,
    # Register your hooks, you can have AFTER_GET, BEFORE_GET
    # you can have multiple hooks, value passes from one to another
    _registered_hooks={
       Action.AFTER_GET: [Hook(get_data_from_somewhere)]
    },
)

# this value can actually come from settings files or envvars
settings.set("token", "@special:/vault/path:token")

assert settings.token == "lets believe it was retrieved from special place"

# from cache this time
assert settings.token == "lets believe it was retrieved from special place"
```

### Merging with access hooks

When the `result.value` is a mergeable data structure like `dict` or `list`
and you want the merging to be applied then the hook must use the `settings`
to wrap the result value and process the merging before returning.

```py
def get_data_from_somewhere(settings, result, key, *args, **kwargs):
    """Your function that goes to external service like gcp, aws, vault, etc
    NOTE: settings in this context is copy of settings, changes made are
    valid only during the scope of this function.
    """
    if key in CACHE:
        return CACHE[key]

    ...

    interesting_keys = ["foo", "bar", "zaz"]  # known to be dicts or lists
    if key in interesting_keys:
      value = CACHE[key] = get_value_from_special_place(key)
      settings.set(key, value)  # process merging and parsing
      return settings.get(key, result.value)

    return result
```

### Registering Hooks After Dynaconf is instantiated

It is ok to register hooks later, which might be useful to do it conditionally.

```py
if something:
    settings.set("_registered_hooks", {Action.AFTER_GET: [Hook(....)]})
```

It also works if you use `dynaconf_hooks.py` to register it conditionally.

```py
# dynaconf_hooks.py
def post(settings):
    data = {}
    if settings.something:
        data["_registered_hooks"] = {Action.AFTER_GET: [Hook(....)]}
    return data
```

### DIFF
--- a/docs/dynamic.md
+++ b/docs/dynamic.md
@@ -183,6 +183,10 @@ NOTES:
 - This may introduce overhead dependending on how the hook function is
     implemented
 
+To register an access hook, you need to wrap the `Dynaconf` instance with
+`HookableSettings` and pass a `Action` to `Hook` mapping in to the constructor
+in the `_registered_hooks` kwarg. For example,
+
 ```python
 # So you have a function that retrieves data from external source
 
--- a/docs/dynamic.md
+++ b/docs/dynamic.md
@@ -237,8 +237,9 @@ assert settings.token == "lets believe it was retrieved from special place"
 ### Merging with access hooks
 
 When the `result.value` is a mergeable data structure like `dict` or `list`
-and you want the merging to be applied then the hook must use the `settings`
-to wrap the result value and process the merging before returning.
+and you want the merging to be applied, then the hook must delegate the merging
+to the given `settings` by `set`ing the result (which will perform the merging)
+and returning the result of `get`ing the setting from the given `settings`.
 
 ```py
 def get_data_from_somewhere(settings, result, key, *args, **kwargs):
--- a/docs/dynamic.md
+++ b/docs/dynamic.md
@@ -177,10 +177,11 @@ cloud provider's secrets managers, or when you need special processing on top
 of raw values (such as decrypting an encrypted value).
 
 NOTES:
-   - You must define the logic of each hook, dynaconf just calls it
-   - You must take care of caching and caching invalidation
-   - This may introduce overhead dependending on how the hook function is
-      implemented
+
+- You must define the logic of each hook, dynaconf just calls it
+- You must take care of caching and caching invalidation
+- This may introduce overhead dependending on how the hook function is
+    implemented
 
 ```python
 # So you have a function that retrieves data from external source
--- a/docs/dynamic.md
+++ b/docs/dynamic.md
@@ -194,8 +194,17 @@ CACHE = {}  # could be Redis
 
 def get_data_from_somewhere(settings, result, key, *args, **kwargs):
     """Your function that goes to external service like gcp, aws, vault, etc
-    NOTE: settings in this context is copy of settings, changes made are
-    valid only during the scope of this function.
+
+    Parameters:
+        settings: A copy of settings, changes made are valid only during the
+            scope of this function.
+        result: The object that holds the setting known to Dynaconf at the
+            moment of access in the .value attribute.
+        key: The accessed key just as passed to Dynaconf.get, usually a
+            single str, but if the requested key was dotted then the whole
+            path is passed to this function.
+        *args and **kwargs: extra positional or named arguments passed to .get
+            such as default,cast,fresh,parent etc.
     """
     if key in CACHE:
         return CACHE[key]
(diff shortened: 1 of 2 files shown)


docs/dynamic.md
```diff
@@ -165,3 +165,116 @@ other_number = "@get number @int 12"
...
+
+# from cache this time
+assert settings.token == "lets believe it was retrieved from special place"
+```
+
+### Merging with access hooks
+
+When the `result.value` is a mergeable data structure like `dict` or `list`
+and you want the merging to be applied then the hook must use the `settings`
+to wrap the result value and process the merging before returning.
```
sebastian-correa, line 238: Non-blocking: I think this is more explicit, though it's non-blocking because I think the example is quite self explanatory.

```suggestion
and you want the merging to be applied, then the hook must delegate the merging
to the given `settings` by `set`ing the result (which will perform the merging)
and returning the result of `get`ing the setting from the given `settings`.
```

sebastian-correa (changes requested):
Thanks for adding this to the docs! I think it's a welcome feature and letting people know about it is always a bonus. Are there any plans about the "privacy" of the hooks attribute? The leading `_` might look daunting to some users.

Not sure what to say about the changes done to `app.py`. Should I just run `make test_functional` and be happy if there are no complains?

BTW, I'd say the only blocking request is the one about the list not rendering properly.

sebastian-correa:
Argh, sorry I reviewed this when @pedro-psb was assigned, I didnt' even notice! My bad! :S

rochacbruno:
Thanks @sebastian-correa I addressed your suggestions

rochacbruno:
@sebastian-correa 

> Are there any plans about the "privacy" of the hooks attribute? The leading _ might look daunting to some users.

It was necessary to avoid conflict for example of a user has a setting variable names `wrapped_class`.

With the new schema on 4.0.0 I think we can make it better by adding the hook individually on each Annotated key.

```py
class Settings(Dynaconf):
    token: Annotated[str, AccessHook(my_function)]
```


> Not sure what to say about the changes done to app.py. Should I just run make test_functional and be happy if there are no complains?

That is just a test that runs on CI, if you run `make tests_functional` or `./tests_functional/runtests.py --filter hooking` it must run and pass the test.

sebastian-correa (approved):
🚀

Thanks for addressing all of my comments and making this new doc section!

PR #1168, opened 2024-08-09
fix: del attr wouldnt work with lowercase

### LINKED ISSUE #1042: [bug] Deleting entry raises an error
**Describe the bug**

As part of my tests, I want to temporarily remove a section of my config to ensure my application reacts properly to this section being missing.

I try `del config.some_section` which produces the following error:

```
In [1]: config
Out[1]: <dynaconf.base.LazySettings object at 0x7fe6597e4c50>

In [2]: config.database
Out[2]: <Box: {'host': 'some_host', 'port': 5431, 'user': 'postgres', 'password': 'secret', 'name': 'the_data'}>

In [3]: del config.database
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 del config.database

File ~/dev/prospect/dataworker/.venv/lib/python3.11/site-packages/dynaconf/utils/functional.py:64, in LazyObject.__delattr__(self, name)
     62 if self._wrapped is empty:
     63     self._setup()
---> 64 delattr(self._wrapped, name)

File ~/dev/prospect/dataworker/.venv/lib/python3.11/site-packages/dynaconf/base.py:300, in Settings.__delattr__(self, name)
    298 self._deleted.add(name)
    299 if hasattr(self, name):
--> 300     super().__delattr__(name)

AttributeError: 'Settings' object has no attribute 'database'

In [4]: config.database
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 config.database

File ~/dev/prospect/dataworker/.venv/lib/python3.11/site-packages/dynaconf/base.py:124, in LazySettings.__getattr__(self, name)
    122     self._setup()
    123 if name in self._wrapped._deleted:  # noqa
--> 124     raise AttributeError(
    125         f"Attribute {name} was deleted, " "or belongs to different env"
    126     )
    128 if name not in RESERVED_ATTRS:
    129     lowercase_mode = self._kwargs.get(
    130         "LOWERCASE_READ_FOR_DYNACONF",
    131         default_settings.LOWERCASE_READ_FOR_DYNACONF,
    132     )

AttributeError: Attribute database was deleted, or belongs to different env
```

**To Reproduce**

Try to delete a section of a dynaconf config object. My config object is not special, I create it using the standard method. It's an instance of `dynaconf.base.LazySettings`. The section in question do exist before I try to delete it.

**Expected behavior**

I expect the command to be successful, not throw an error, and subsequent attempts to access this section to fail as if it had never been configured in the first place.

 - Dynaconf Version: 3.2.4

**Comments**

It looks that there is some code written to handle deleting a section of the config but

pedro-psb:
Closes #1042 

Notes:
* The first level is always uppercase internally and calling del with different case attr-name would fail.
* Currently we only support deleting a nested level item with exact matching (not sure if needs "fixing").

### DIFF
--- a/dynaconf/utils/functional.py
+++ b/dynaconf/utils/functional.py
@@ -63,7 +63,7 @@ class LazyObject:
             raise TypeError(f"can't delete {name}.")
         if self._wrapped is empty:
             self._setup()
-        delattr(self._wrapped, name)
+        delattr(self._wrapped, name.upper())
 
     def _setup(self):
         """
--- a/dynaconf/vendor/box/box.py
+++ b/dynaconf/vendor/box/box.py
@@ -15,6 +15,7 @@ from pathlib import Path
 from typing import Any, Union, Tuple, List, Dict
 
 from dynaconf.vendor import box
+from dynaconf.utils import find_the_correct_casing
 from .converters import (_to_json, _from_json, _from_toml, _to_toml, _from_yaml, _to_yaml, BOX_PARAMETERS)
 from .exceptions import BoxError, BoxKeyError, BoxTypeError, BoxValueError, BoxWarning
 
@@ -402,7 +403,7 @@ class Box(dict):
                     if _camel_killer(key) == each_key:
                         key = each_key
                         break
-        super().__delitem__(key)
+        super().__delitem__(find_the_correct_casing(key, self))
 
     def __delattr__(self, item):
         if self._box_config['frozen_box']:
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -22,7 +22,7 @@ pytest
 pytest-cov
 pytest-mock
 commentjson
-lovely-pytest-docker
+pytest-docker
 tox
 
 # style check
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -25,6 +25,26 @@ def test_deleted_raise(settings):
     assert settings.exists("TODELETE") is False
     assert settings.get("TODELETE") is None
 
+    # case: del using lowercase when original was mixed-case
+    settings.ToDelete3 = True
+    assert "ToDelete3" in settings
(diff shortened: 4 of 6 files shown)


dynaconf/utils/functional.py
```diff
@@ -63,7 +63,7 @@ def __delattr__(self, name):
             raise TypeError(f"can't delete {name}.")
         if self._wrapped is empty:
             self._setup()
-        delattr(self._wrapped, name)
+        delattr(self._wrapped, name.upper())
```
rochacbruno, line 66: Cant this use https://github.com/dynaconf/dynaconf/blob/master/dynaconf/utils/__init__.py#L519-L540 ?

to find the correct casing?

Well, when we  move to schema based this will not matter anymore

rochacbruno:
we may need to force docker-compose, or change the referenes to call `docker-compose`

```yaml
- name: (Linux) Install docker compose
  run: |
    curl -L -o /tmp/docker-compose https://github.com/docker/compose/releases/download/v2.29.1/docker-compose-linux-x86_64
    install /tmp/docker-compose /usr/local/bin/
```

dynaconf/utils/functional.py
pedro-psb: Possibly, I'll give it a try

pedro-psb:
This raw patch on lovely-pytest-docker fixes the issue: https://github.com/pedro-psb/lovely-pytest-docker/commit/fce231178618f1e2e729d6c355c0d38fdbb02d39

But looks like it's brother https://github.com/avast/pytest-docker is slighly more maintained and uses docker compose by default, so I think we should switch to it.

rochacbruno (approved):

PR #1204, opened 2024-12-11
AWX Integration fixes

rochacbruno:
This PR contains fixes and features for Ansible AWX integration.

Some are backport from master (-json on list CLI)

Some are features that will need to be ported to master
- multiple envvar_prefix
- composed current_env
- More tracking sourcemetadata
- `@insert` token

DOCS will be added in a follow up PR against master branch #1224


dynaconf/cli.py
```diff
@@ -117,6 +113,18 @@ def set_settings(ctx, instance=None):
...
+    try:
+        with redirect_stdout(None):
+            module = importlib.import_module(django_settings_module)
+    except ImportError as e:
+        raise click.UsageError(e)
+    except FileNotFoundError:
+        return
+    return module
+
+
```
pedro-psb, line 127: TIL `with redirect_stdout`

dynaconf/loaders/env_loader.py
```diff
@@ -66,7 +82,7 @@ def load_from_env(
         env_ = f"{prefix}_"
 
     # set source metadata
-    source_metadata = SourceMetadata(identifier, "unique", "global")
+    source_metadata = SourceMetadata(identifier, prefix, "global")
```
pedro-psb, line 85: Nice, it will show from what prefix it came from!

dynaconf/utils/__init__.py
```diff
@@ -78,6 +79,10 @@ def safe_items(data):
             else:
                 return data.items()
 
+        # TODO @rochacbruno: Make the above line work with 2 dynaconf_merge
+        # https://github.com/dynaconf/dynaconf/issues/todo
+        # {"a": ["d", "e"], "dynaconf_merge": True},
+
```
pedro-psb, line 85: Is this done or it will be done later?

dynaconf/utils/parse_conf.py
```diff
@@ -141,6 +141,52 @@ def __init__(self, value, box_settings, unique=False):
...
+            42 -> index = 0, value = 42
+            42 foo -> index = 42, value = foo
+            42 foo bar -> index = 42, value = foo bar
+        """
+        self.box_settings = box_settings
+
+        try:
+            if value[0].isdigit():
+                # `0 foo` or `-1 foo` or `42 foo` or `42`(raise ValueError)
+                index, value = value.split(" ", 1)
```
pedro-psb, line 176: This won't be true for `value == "-1 foo"` because `value[0] == "-"`.

tests/test_base.py
pedro-psb, line 1: :tada:

tests/test_settings_loader.py
```diff
@@ -0,0 +1,167 @@
...
+    settings.setenv("PRODUCTION,special")
+    dummye_folder = tmpdir.mkdir("dummye")
+
+    dummye_folder.join("dummye_module.py").write('FOOB = "bar"')
+    dummye_folder.join("__init__.py").write('print("initing dummye...")')
+
+    dummye_folder.join("production_dummye_module.py").write('BAZ = "zaz"')
+    dummye_folder.join("special_dummye_module.py").write("DEBUG = False")
+
+    settings_loader(settings, "dummye.dummye_module")
```
pedro-psb, line 139: So if `environments=True` the `settings_loader` will interpret any prefix in the module name as an env, right?

tests/test_settings_loader.py
```diff
@@ -0,0 +1,167 @@
...
+    assert "PRODUCTION" in settings.current_env
+    assert settings.BAZ == "zaz"
+    assert settings.DEBUG is False  # comes from special_
+
+
+def test_load_using_settings_loader_with_one_env_named_file_file_path_multi_env(
+    tmpdir,
+):
+    settings = Dynaconf(environments=True)
+    settings.setenv("PRODUCTION,special")
```
pedro-psb, line 152: What exactly does it mean or is the effect of doing `settings.setenv` with two envs?

tests_functional/insert/app.py
pedro-psb, line 1: Is there a way to insert a dictionary?
From what I understand from the implementation, the following wouldn't work, right?
`'@insert 0 @json { ... }'`

pedro-psb (approved):
Nice! It looks all reasonable to me.
Most comments were so I could be more up-to-date with the new behavior.
Assuming I got the `@insert -1` error right, it's up to you. Do you wanna fix/support it now?

tests/test_settings_loader.py
pedro-psb: (I read the PR description just now, that's the "composed current_env" then).

And how is it different from the init kwargs `environments` list?

dynaconf/utils/__init__.py
rochacbruno: I will handle it later

dynaconf/utils/__init__.py
```diff
@@ -432,7 +451,11 @@ def _recursively_evaluate_lazy_format(
         value = value(settings)
 
     if isinstance(value, list):
-        # Keep the original type, can be a BoxList
+        # This must be the right way of doing it, but breaks validators
+        # To be changed on 4.0.0
+        # for idx, item in enumerate(value):
+        #     value[idx] = _recursively_evaluate_lazy_format(item, settings)
+
```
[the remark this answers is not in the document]
rochacbruno: Yes, I cannot fix this right now because changing it opens a can of worms, I want to completely change the validator behavior and that will be a breaking change

dynaconf/utils/parse_conf.py
rochacbruno: Good catch, I made this way to avoid regexp but now I need to rethink

tests/test_settings_loader.py
rochacbruno: Yes, this was actually the first implementation of multiple envs, I would like it to be the reverse `{module}_{env}.py` but now it is too late to change

tests_functional/insert/app.py
rochacbruno: It will work because the value will be "@json ..." and this will be passed to `parse_conf_data`

pedro-psb: Cool

dynaconf/utils/parse_conf.py
pedro-psb: Do you wanna do this on this PR or can I merge it?

rochacbruno: I will work on a fix right now and update the PR

rochacbruno: Fixed!

PR #1246, opened 2025-02-11
feat: Add support for decorated hooks on settings files

rochacbruno:
Related to #1245
To be ported to master branch after merge

#### Decorator approach

**new in 3.2.8**

When the settings file is a Python file, you can also use a decorator to define hooks.

```python
from dynaconf import Dynaconf

settings = Dynaconf(settings_file="settings.py")
```

`settings.py`
```python
from dynaconf import post_hook

VARIABLE = "value"
# all regular settings here
...

@post_hook
def set_debugging_database_url(settings):
    data = {}
    if settings.DEBUG:
        data["DATABASE_URL"] = "sqlite://"
    return data
```

Dynaconf will collect the decorated functions and execute them after the settings file is loaded.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -81,6 +81,37 @@ settings = Dynaconf(post_hooks=hook_function)
 You can also set the merging individually for each settings variable as seen on
 [merging](merging.md) documentation.
 
+
+#### Decorator approach
+
+**new in 3.2.8**
+
+When the settings file is a Python file, you can also use a decorator to define hooks.
+
+```python
+from dynaconf import Dynaconf
+
+settings = Dynaconf(settings_file="settings.py")
+```
+
+`settings.py`
+```python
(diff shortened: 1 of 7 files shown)


docs/advanced.md
```diff
@@ -81,6 +81,37 @@ settings = Dynaconf(post_hooks=hook_function)
...
+When the settings file is a Python file, you can also use a decorator to define hooks.
+
+```python
+from dynaconf import Dynaconf
+
+settings = Dynaconf(settings_file="settings.py")
+```
+
+`settings.py`
+```python
```
pedro-psb, line 98: This displays better ([example](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-a-title))
```suggestion
```python title="settings.py"
```

pedro-psb (approved):

PR #1274, opened 2025-04-02
fix: Fix merge_unique stability (keep elements order)

pedro-psb:
WIP: fix the test for the intended behavior

Closes #1167


pedro-psb:
We are leaving this for 4.0.0, as it'll break tested behavior

PR #1275, opened 2025-04-04
Run CI test against distribution wheel

pedro-psb:
This PR:
* Adds missing build/release dependencies: `python-minifier` and `build`.
* Makes CI use (as installation source) the distribution wheel we expect to release
* Adds check to the `minify.sh` script to ensure we are using our lowerbound python, as it may create incompatible changes to the vendored code otherwise. (as reported by Bruno!)

### DIFF
--- a/.github/actions/install-from-wheel/action.yml
+++ b/.github/actions/install-from-wheel/action.yml
@@ -1,0 +1,55 @@
+name: 'Install from wheel'
+description: 'Setup Python and install Dynaconf wheel with optional dependencies'
+
+inputs:
+  dependencies:
+    description: 'Optional dependencies to install with the wheel'
+    required: false
+    default: ''
+  os:
+    description: 'Operating system (Windows, Linux, macOS)'
+    required: false
+    default: 'Linux'
+  python-version:
+    description: 'Python version to use'
+    required: false
+    default: '3.9'
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/setup-python@v5
+      with:
+        python-version: ${{ inputs.python-version }}
+    - name: Upgrade PIP
+      shell: bash
+      run: pip install --user --upgrade pip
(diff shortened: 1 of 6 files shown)


pedro-psb:
The commits may be rebased instead of squashed.

.github/actions/install-from-wheel/action.yml
```diff
@@ -0,0 +1,55 @@
...
+
+inputs:
+  dependencies:
+    description: 'Optional dependencies to install with the wheel'
+    required: false
+    default: ''
+  os:
+    description: 'Operating system (Windows, Linux, macOS)'
+    required: false
+    default: 'lnux'
```
rochacbruno, line 12: is this a typo?
```suggestion
    default: 'linux'
```

.github/actions/install-from-wheel/action.yml
```diff
@@ -0,0 +1,55 @@
...
+      continue-on-error: ${{ inputs.os == 'Windows' }}
+
+    - name: Download wheel
+      uses: actions/download-artifact@v4
+      with:
+        name: dynaconf_dist
+        path: dist/
+
+    - name: Install from wheel (Unix)
+      if: inputs.os != 'Windows'
```
rochacbruno, line 36: gotcha! typo on os doesn't make difference in this case

rochacbruno (approved):

.github/workflows/main.yml
```diff
@@ -29,24 +29,42 @@ concurrency:
...
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.9"
       - name: Upgrade PIP
         run: pip install --user --upgrade pip
       - name: Install and lint
         run: make clean install run-pre-commit
 
-  install_test:
+  build:
```
mdellweg, line 43: Loving the idea to move the PR tests as close as possible to the actual release process!

PR #1380, opened 2026-06-12
Fix index merge padding scalar lists with nested empty lists

sarathfrancis90:
With `INDEX_SEPARATOR_FOR_DYNACONF` enabled, setting an index of a plain list of scalars fills the skipped positions with empty lists instead of leaving them empty. For example:

```python
settings = Dynaconf(INDEX_SEPARATOR_FOR_DYNACONF="___")
settings.set("ports[2]", 9090)
# settings.PORTS == [[], [], 9090]   # expected [None, None, 9090]
```

Same via env vars (`DYNACONF_PORTS___2=9090`). You end up with nested empty lists where you only wanted to assign one element.

The padding in `_dotted_set` reused `next_default`, which is only recomputed for non-final path segments, so on the leaf index it still referenced the parent level's container default. I pad the gaps with `None` when the index is the last segment; nested (non-leaf) lists are untouched and still pad with their proper container type, e.g. `matrix[1][1]` stays `[[], [None, "x"]]`.

Added two tests under `TestIndexMerge` covering the `.set` and dunder-env paths.

### DIFF
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -1086,10 +1086,17 @@ class Settings:
                     and isinstance(tree, list)
                 ):  # accessing index of a list
                     index = int(k.replace("[", "").replace("]", ""))
-                    extended_list = [
-                        next_default.copy()
-                        for _ in range(index + 1)  # type: ignore[attr-defined]
-                    ]
+                    # Pad missing positions so we can assign any index.
+                    # On the last segment the slot holds the value itself, so
+                    # gaps should be empty (None) instead of a copy of
+                    # `next_default`, which still refers to the previous level.
+                    if is_not_end:
+                        extended_list = [
+                            next_default.copy()  # type: ignore[attr-defined]
+                            for _ in range(index + 1)
+                        ]
+                    else:
+                        extended_list = [None] * (index + 1)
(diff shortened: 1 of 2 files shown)


pedro-psb (commented):
Can you share how did you find out about this? It's unreleased code.

sarathfrancis90:
I was looking into how dynaconf handles list index assignment and found the `INDEX_SEPARATOR_FOR_DYNACONF` path on main, so I tried it out. Setting a sparse index on a plain scalar list — `settings.set("ports[2]", 9090)` (and the same via `DYNACONF_PORTS___2`) — gave me `[[], [], 9090]` where I expected `[None, None, 9090]`. That surprised me enough to dig in, and it traced back to the gap-padding in `_dotted_set` reusing the parent level's `next_default` on the leaf index. I've been running from main since it isn't in a release yet.

pedro-psb (approved):
The behavior of creating "empty" entries (list, dicts or nones) bothers me a bit, but it's an improvement in any case.

Maybe we should have a more simple rule, like simply behaving like an `list.insert` (which accepts indexes out of range and appends to the end) or like list set/replace. But it's in tech preview, so lets wait for feedback.

PR #1383, opened 2026-06-19
fix: prevent RecursionError on dotted set with bracket in first segment

sarathfrancis90:
`settings.set("a[1]", "x")` blows up with `RecursionError: maximum recursion depth exceeded`. Same thing for any dotted key whose first segment carries a bracket, e.g. `settings.set("servers[0].name", "web1")`. This is with index merge disabled, which is the default (`INDEX_SEPARATOR_FOR_DYNACONF` unset).

With index merge off a bracket is just part of a literal key — `test_dotted_set_with_index_merge_disabled` already documents that `nested_a.nested_b[2][1]...` stores `nested_b[2][1]` as a plain key. The crash only happens when the bracket lands in the first segment: `_dotted_set` writes the resolved tree back through `self.update`, and the top-level key it produces (`split_keys[0]`) still contains `[`, so `update` routes it straight back into `_dotted_set` (any `[` is treated as a dotted path) and it loops.

I pass `dotted_lookup=False` on that final `update` so the already-resolved keys are written literally instead of being re-parsed. Non-first-segment literal keys and the index-merge path are unchanged.

Added a regression test in `TestIndexMerge`; it raises RecursionError before the change and passes after.

### DIFF
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -1107,11 +1107,17 @@ class Settings:
                 full_path=split_keys,
                 list_merge=list_merge,  # when to use deep / shallow replace?
             )
+        # `new_data` is keyed by the already-resolved top level key
+        # (`split_keys[0]`). With index merge disabled a bracket is a literal
+        # key, so that key can still contain `[` (e.g. `servers[0]`). Writing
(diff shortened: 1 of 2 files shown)


pedro-psb (approved):
Looks good, thanks for the contribution.

PR #1392, opened 2026-06-24
Avoid evaluating lazy values during env overrides

sjh9714:
## Summary

Fixes #1356.

Environment overrides should be able to replace a lazy `@jinja` value without evaluating the old value first. `Settings.set()` was using the normal store getter for merge bookkeeping, which could evaluate the existing lazy value before the replacement was stored and raise `UndefinedError` for nested include references that were no longer needed.

## Changes

- Use the raw store lookup for the existing value during `Settings.set()` bookkeeping so plain overrides do not evaluate stale lazy values.
- Add regression coverage for overriding a nested `@jinja` value loaded through `dynaconf_include`.

## Testing

- `uv run pytest tests/test_base.py::test_env_override_nested_lazy_value_from_included_file -q`
- `uv run ruff check dynaconf/base.py tests/test_base.py`
- `LOAD_DOTENV_FOR_DYNACONF=true DOTENV_PATH_FOR_DYNACONF=tests/.env uv run pytest tests/test_base.py -q`

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.

### DIFF
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -1200,10 +1200,12 @@ class Settings:
             tomlfy_filter=tomlfy_filter,
         )
 
-        # Fix for #869 - The call to getattr trigger early evaluation
-        existing = (
-            self.store.get(key, None) if not isinstance(parsed, Lazy) else None
-        )
+        # Fix for #869 - Evaluating an existing lazy value during set can
+        # fail before a complete replacement value is stored.
+        existing = None
+        if not isinstance(parsed, Lazy):
+            with suppress(AttributeError, KeyError):
+                existing = self.store.get(key, bypass_eval=True)
 
         if getattr(parsed, "_dynaconf_insert", False):
             # `@insert` calls insert in a list by index
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -1205,7 +1205,10 @@ class Settings:
         existing = None
         if not isinstance(parsed, Lazy):
             with suppress(AttributeError, KeyError):
-                existing = self.store.get(key, bypass_eval=True)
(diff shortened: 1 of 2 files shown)


pedro-psb (changes requested):
Hello @sjh9714 ,
This introduces some regressions.
You can test locally with make test_all .

sjh9714:
Thanks, I pushed a follow-up in bdba184.

The regression came from calling `self.store.get(..., bypass_eval=True)` even when `DYNABOXIFY=False` uses a plain `dict` store, such as Flask compatibility mode. The update keeps the `bypass_eval=True` path for `DataDict` and falls back to normal `dict.get()` for plain dict stores.

Local checks run:

- `uv run pytest tests/test_flask.py::test_compatibility_mode -q`
- `uv run pytest tests/test_base.py::test_env_override_nested_lazy_value_from_included_file -q`
- `uv run pytest tests/test_flask.py::test_compatibility_mode tests/test_base.py::test_env_override_nested_lazy_value_from_included_file -q`
- `uv run pytest tests/test_base.py -k "env_override_nested_lazy_value_from_included_file or dotted_set or from_env_method" -q`
- `uv run python -m py_compile dynaconf/base.py`
- `git diff --check`

I also tried `uv run make test_all`; it progressed through many functional tests but stopped in `tests_functional/legacy/django_pytest` with a local sqlite setup failure: `no such table: auth_user`. The new GitHub Actions run is currently `action_required`, so it looks like it may need maintainer approval before CI starts.

pedro-psb (approved):
Thanks for the contribution.

PR #1404, opened 2026-06-25
fix: Remove call to deprecated functions on public APIs

### LINKED ISSUE #1393: [bug] 3.3.0 deprecates an interface in use by Dynaconf
**Describe the bug**
`Dynaconf().as_dict()`  calls `self.store.to_dict()`  which is deprecated in notes.py. This just broke my CICD pipeline as the deprecation warnings are errors.

**To Reproduce**
Call Dynaconf().as_dict() 

**Expected behavior**
Supported APIs don't raise deprecation warnings.

 - Dynaconf Version 3.3.0

pedro-psb:
* Add fixture to catch that our tests using public API don't emit deprecation themselves
* Fix deprecations being called for indirect deprecated method calls

Closes: #1393

### DIFF
--- a/dynaconf/nodes.py
+++ b/dynaconf/nodes.py
@@ -426,6 +426,7 @@ class DataList(list):
         box_deprecation_warning(
             "to_list", "DataList", "Use list(data_list) instead."
         )
+        # to_dict recursively converts to both dicts and lists
         return ut.to_dict(self)
 
     def to_json(
--- a/dynaconf/nodes.py
+++ b/dynaconf/nodes.py
@@ -171,15 +171,7 @@ class DataDict(dict):
         box_deprecation_warning(
             "to_dict", "DataDict", "Use dict(data_dict) instead."
         )  # pragma: nocover
-        out_dict = dict(self)
-        for k, v in out_dict.items():
-            if v is self:
-                out_dict[k] = out_dict
-            elif isinstance(v, DataDict):
-                out_dict[k] = v.to_dict()
-            elif isinstance(v, DataList):
-                out_dict[k] = v.to_list()
-        return out_dict
+        return ut.to_dict(self)
 
     def merge_update(self, __m=None, **kwargs):  # pragma: nocover
         """Merge update with another dict"""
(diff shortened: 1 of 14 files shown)


rochacbruno:
`write()` still calls `DataDict.to_dict()` https://github.com/dynaconf/dynaconf/blob/704099c179ad8da739dd4414d0f914f9410377e8/dynaconf/loaders/__init__.py#L446 

`dynaconf list --output` would emit the warning.

I know we talked about deprecating write method but that is for >4 only I guess.

tests/test_base.py
```diff
@@ -151,15 +154,12 @@ class Obj:
...
     assert isinstance(obj.ALIST, DataList)
-    assert isinstance(obj.ADICT.to_yaml(), str)
 
     # now make sure convert_to_dict=True brings in dict and list
     obj = Obj()
     settings.populate_obj(obj, convert_to_dict=True)
     assert isinstance(obj.ADICT, dict)
     assert isinstance(obj.ALIST, list)
-    with pytest.raises(AttributeError):
-        assert isinstance(obj.ADICT.to_yaml(), str)
```
rochacbruno, line 163: I think we need to add an `assert type(obj.ADICT) is dict` to cover the case you mentioned?

tests/conftest.py
```diff
@@ -11,6 +12,13 @@
 from dynaconf.base import LazySettings
 
 
+@pytest.fixture
+def no_deprecations():
+    with warnings.catch_warnings():
+        warnings.simplefilter("error", DeprecationWarning)
+        yield
```
rochacbruno, line 21: this is catching all the DeprecationWarning from all the libs, even stdlib DeprecationWarning would be filtered out.

Filtering with `warnings.filterwarnings('error', category=DeprecationWarning, module='dynaconf')` would make it specific.

dynaconf/nodes.py
```diff
@@ -457,17 +451,7 @@ def to_list(self):  # pragma: nocover
...
-            if x is self:
-                new_list.append(new_list)
-            elif isinstance(x, DataDict):
-                new_list.append(x.to_dict())
-            elif isinstance(x, DataList):
-                new_list.append(x.to_list())
-            else:
-                new_list.append(x)
-        return new_list
+        return ut.to_dict(self)
```
rochacbruno, line 430: the main outer function is named `to_list` but it is returning a `ut.to_dict`, is that right?

tests/test_nodes.py
```diff
@@ -433,7 +433,7 @@ def test_box_list(self, tmp_path, deprecated_context):
             method = getattr(DataList, method_name)
             assert method(filename=filename) == test_data
 
-    def test_box_list_copying(self, deprecated_context):
+    def test_box_list_copying(self):
```
rochacbruno, line 452: This removes the deprecated_content fixture, but the no_deprecation is not loaded anywhere here in this module, if someone re-adds the deprecation warning to the methods we would not catch the regression.

Maybe marking the no_deprecation module fixture here helps?

rochacbruno (dismissed):
Approve.

Added few minor comments, feel free to not address those if you think it is ok.

tests/test_base.py
pedro-psb: Oh, that's right! `isinstance` is not telling what we want.

dynaconf/nodes.py
pedro-psb: The `to_dict` utility converts list-like objects to list too,  which is equivalent.
But yeah, the name is a bit misleading...

rochacbruno: maybe adding a comment.

PR #1414, opened 2026-06-30
fix(py_loader): re-raise ImportError caused by module content, not just missing modules

uttam12331:
When loading a settings module by name, `get_module()` calls
`importlib.import_module()` and catches `ImportError` broadly to fall back
to file-based loading. This has a subtle flaw: any `ImportError` raised
*inside* the module body (cyclic imports, missing transitive dependencies,
bad `from x import y` statements) is also caught and silently swallowed,
even when `silent=False`.

The caller ends up with `(None, None)` instead of an exception, so
dynaconf silently loads no settings and the user has no indication
anything went wrong.

**Root cause**

```python
# before
except (ImportError, TypeError):   # catches ALL ImportErrors
    mod = import_from_filename(...)
```

The `ImportError` catch is only needed to distinguish "this name is not an
importable module" from "this name is a file path". Python 3.6 introduced
`ModuleNotFoundError` (a subclass of `ImportError`) which is raised
*only* when the module itself cannot be located — not when an error occurs
inside it. Using it here preserves the fallback behaviour for missing
modules while letting real import errors propagate.

**Fix**

```python
# after
except (ModuleNotFoundError, TypeError):   # only catches "module not found"
    mod = import_from_filename(...)
```

`ModuleNotFoundError` is available on Python 3.6+ (dynaconf already
requires Python 3.8+), so there is no compatibility concern.

A regression test is added to `tests/test_py_loader.py` that patches
`importlib.import_module` to raise a plain `ImportError` (simulating a
cyclic-import inside the settings file) and asserts it propagates when
`silent=False`.

Closes #1308

### DIFF
--- a/tests/test_py_loader.py
+++ b/tests/test_py_loader.py
@@ -1,11 +1,14 @@
 from __future__ import annotations
 
 import os
+import sys
+import types
 
 import pytest
 
 from dynaconf import default_settings
 from dynaconf import LazySettings
+from dynaconf.loaders.py_loader import get_module
 from dynaconf.loaders.py_loader import load
 from dynaconf.loaders.py_loader import try_to_load_from_py_module_name
 from dynaconf.utils import DynaconfDict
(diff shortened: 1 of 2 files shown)


tests/test_py_loader.py
pedro-psb, line 1: Can you use the `create_file` fixture to actually create modules with the cyclic import setup?
I'd rather test against an actual setup rather than relying on internals details of the problem.

pedro-psb (changes requested):

uttam12331:
Updated the test to use real on-disk modules via the `create_file` fixture (see latest commit). The test now:

1. Creates `cyclic_helper.py` with a defined name (`READY`)
2. Creates `cyclic_settings.py` that tries to import a name that doesn't exist from it (`DOES_NOT_EXIST`) — the exact error Python raises in a real cyclic-import scenario
3. Temporarily adds `tmp_path` to `sys.path` so the module is importable, then asserts `ImportError` is raised

No more `mock.patch` or `types.ModuleType` — it exercises the real import path.

tests/test_py_loader.py
```diff
@@ -275,3 +278,39 @@ def test_post_load_hooks(clean_env, tmpdir):
...
+    settings = DynaconfDict()
+
+    sys.path.insert(0, str(tmp_path))
+    try:
+        with pytest.raises(ImportError):
+            get_module(settings, "cyclic_settings", silent=False)
+    finally:
+        sys.path.remove(str(tmp_path))
+        for mod in ("cyclic_settings", "cyclic_helper"):
+            sys.modules.pop(mod, None)
```
pedro-psb: Thanks, this is much better.
It's fine to test this scenario of a missing name, but it should also cover the main problem setup, which is a cyclic import. This doesn't have a cyclic import.

Also it would be nice to extract this `sys.path` include to a fixture like `with_importable_tmp_path`.

pedro-psb (changes requested):

tests/test_py_loader.py
pedro-psb: Also, please run `make run-pre-commit` to fix linting errors before pushing.

uttam12331:
Updated again per your feedback:

1. **Real cyclic import**: `cyclic_a.py` imports from `cyclic_b`, and `cyclic_b` imports from `cyclic_a` — a genuine A→B→A cycle. Python raises `ImportError: cannot import name 'A_VALUE' from partially initialized module 'cyclic_a'`.
2. **`with_importable_tmp_path` fixture**: extracted the `sys.path` insertion/cleanup and module eviction into a reusable fixture.
3. **Pre-commit**: ran `ruff format` + `ruff check` — fixed formatting and removed an unused import.

pedro-psb (changes requested):
I ran the test without the patch and it passes, so this is not reproducing the problem. The import error may be related to something else. Also, the raises assertion should contain a match for the error message we expect to make sure we are testing the right thing.

Can you confirm the problem the user reports happens outside the testing framework? If yes, there is probably something wrong with the fixture setup.

PR #1421, opened 2026-07-04
Support list values for env (#1278)

### LINKED ISSUE #1278: set `env` with list of two environments raised error: List has no attribute upper
I've encountered this error when I was:

```python
        self._configs = Dynaconf(
            settings_files=[self.config_folder / '*.toml'],
            validators=self._validators,
            environments=True,
            env=['test', 'dev'],
            envvar_prefix='ATM',
            env_switcher='ATM_ENV',
        )
```
The error was:

  File "<string>", line 1, in <module>
  File "/home/demo/Live-usb-storage/projects/python/demo/lib/python3.12/site-packages/dynaconf/base.py", line 122, in __getattr__
    self._setup()
  File "/home/demo/Live-usb-storage/projects/python/demo/lib/python3.12/site-packages/dynaconf/base.py", line 183, in _setup
    self._wrapped = self._wrapper_class(
                    ^^^^^^^^^^^^^^^^^^^^
  File "/home/demo/Live-usb-storage/projects/python/demo/lib/python3.12/site-packages/dynaconf/base.py", line 273, in __init__
    self.execute_loaders()
  File "/home/demo/Live-usb-storage/projects/python/demo/lib/python3.12/site-packages/dynaconf/base.py", line 1184, in execute_loaders
    env = (env or self.current_env).upper()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

setting `env` to string `'test, dev'` would not work either as dynaconf picked it up as env called 'test,dev'. 

I am very new to dynaconf so I didn't raise this as an bug in case I missed something. 

I read this as well in the documentation:

- The env can be a comma separated list of envs, ex: export ENV_FOR_DYNACONF=production,staging will activate both production and staging environments, and variables from both layers will be loaded in the order they are defined.

Does "will activate both environments" indicate that both settings are going to be available during runtime and we can switch `env` by `setenv`, `from_env`, `using_env`?

As a new user to dynaconf , I was under the impression that by using the constructor with `environments=True`, I'd be loaded with environments with custom names I set in my `toml` files, i.e., `[dev.some.value]` and `[test.some.value]` would be picked up correctly as two envs called `dev` and 'test'. 

However, without setting both `environments` and `env` together, the environment picked up got defaulted to `DEVELOPMENT` which was the default by dynaconf. Under `env` we have this line, 

"When environments is set, this controls the current environment used in runtime."

Probably we should rephrase it to that `env` has to be set if environments is True if your have custom environment names.

apoorvdarshan:
Fixes #1278.

## Problem

Dynaconf already supports activating multiple environments through a comma-separated `ENV_FOR_DYNACONF` value, but the equivalent constructor form failed:

```python
Dynaconf(environments=True, env=["test", "dev"])
# AttributeError: 'DataList' object has no attribute 'upper'
```

As discussed in #1278, the list form should be the programmatic equivalent of `ENV_FOR_DYNACONF=test,dev`: load and merge `test`, then `dev`.

## Fix

Normalize a list supplied through either `env` or `ENV_FOR_DYNACONF` to the existing comma-separated representation in `normalize_kwargs()`. This lets all existing multi-environment loader behavior remain unchanged. Lists containing non-string items still raise a clear `TypeError`.

## Testing

- Added regression coverage for both constructor aliases.
- Verified values from both environments are loaded and the later environment wins for overlapping scalar values.
- Core suite: `748 passed, 9 skipped, 19 deselected, 1 xfailed`.
- Focused tests pass on Python 3.10, 3.13, and 3.14.
- Ruff check and format checks pass.

---

*Disclosure: this change was prepared with AI assistance. I reproduced the issue, implemented and verified the fix and tests, and take responsibility for the contribution and review follow-up.*

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -307,9 +307,10 @@ def test_env_kwarg_accepts_a_list(create_file, env_kwarg):
     assert settings.issue_1278_from_dev is True
 
 
-def test_env_kwarg_rejects_a_list_with_non_strings():
+@pytest.mark.parametrize("env", [["test", 1], 1])
+def test_env_kwarg_rejects_invalid_values(env):
     with pytest.raises(TypeError, match="string or a list of strings"):
-        Dynaconf(environments=True, env=["test", 1])
+        Dynaconf(environments=True, env=env)
 
 
 def test_env_should_allow_underline(settings):
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -277,6 +277,21 @@ def test_env_should_be_string(settings):
         settings.setenv(123456)
 
 
+def test_env_kwarg_must_be_a_string():
+    # A non-string ``env`` (e.g. a list) used to fail later with a cryptic
+    # ``'DataList' object has no attribute 'upper'``. See #1278.
+    with pytest.raises(TypeError, match="'env' must be a string"):
+        Dynaconf(environments=True, env=["test", "dev"])
+    # The ENV_FOR_DYNACONF alias is validated too.
+    with pytest.raises(TypeError, match="'env' must be a string"):
+        Dynaconf(environments=True, ENV_FOR_DYNACONF=["test"])
+    # A string env still works.
+    assert (
+        Dynaconf(environments=True, env="production").current_env
+        == "production"
+    )
+
+
 def test_env_should_allow_underline(settings):
     settings.setenv("COOL_env")
     assert settings.current_env == "COOL_ENV"
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -282,17 +282,17 @@ def test_env_kwarg_accepts_a_list(create_file, env_kwarg):
     settings_file = create_file(
         "settings.toml",
         """
-[default]
-issue_1278_value = "default"
+        [default]
+        issue_1278_value = "default"
 
-[test]
-issue_1278_value = "test"
-issue_1278_from_test = true
+        [test]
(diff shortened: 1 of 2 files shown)


pedro-psb (changes requested):
There is some new discussion in the issue, and this is not the expected behavior.
Do you wanna proceed with the new approach discussed  there or can I close the PR?

tests/test_base.py
```diff
@@ -277,6 +277,41 @@ def test_env_should_be_string(settings):
...
+
+[test]
+issue_1278_value = "test"
+issue_1278_from_test = true
+
+[dev]
+issue_1278_value = "dev"
+issue_1278_from_dev = true
+"""
+    )
```
pedro-psb, line 296: Please use the `create_file` fixture.

pedro-psb (changes requested):

tests/test_base.py
```diff
@@ -277,6 +277,41 @@ def test_env_should_be_string(settings):
...
+[test]
+issue_1278_value = "test"
+issue_1278_from_test = true
+
+[dev]
+issue_1278_value = "dev"
+issue_1278_from_dev = true
+""",
+    )
+
```
pedro-psb, line 297: Using `create_file` means not requiring the indentation.
Please unindent the block.

pedro-psb (changes requested):

pedro-psb (approved):

Issue #1187, opened 2024-10-17
[bug] Using get_fresh with dot separated variable is broken


omri-cavnue:
**Describe the bug**
Using the Fresh getter with any dot separated key always returns None after the initial call and fails to load from external loaders. For example, if your settings file consists of:
```bash
[foo]
bar = "baz"
```
Then calling `settings.get("foo.bar", fresh=True) will return "baz" the first time, but then return None any future call. The reason for this is that within the get method on the Settings class, the recursive `self._dotted_get` function splits the parent and child so that on the first call to `self._get` from within `self._dotted_get`, it reaches the `if fresh` line and tries to load the value solely based on the parent key and not does not include any of the children.

**To Reproduce**

1. Having the following folder structure

```bash
config.py
settings.toml
test.py
```

2. Having the following config files:

```bash
from dynaconf import Dynaconf

settings = Dynaconf(
    envvar_prefix="DYNACONF",
    settings_files=["settings.toml"],
    redis_enabled=True,
    apply_default_on_none=False,
)
```

**/path/settings.toml**
```toml
[foo]
bar="baz"
```

3. Having the following app code:

Using `dynaconf -i config.settings  write redis -v foo.bar=test`

**/path/src/app.py**
```python
import time
from dynaconf.loaders import redis_loader
from config import settings
from redis import StrictRedis

redis = StrictRedis(host='localhost', port=6379)


pubsub = redis.pubsub()
pubsub.psubscribe('__keyspace@0__:DYNACONF_MAIN')

try:
    while True:
        message = pubsub.get_message()
        if message and message.get("pattern", None):
           print("NEW", settings.get_fresh("foo.bar")
            
        print("CURRENT", settings.get("foo.bar", fresh=True))
        time.sleep(1)
except:
    print("EXITING")
    redis_loader.delete(settings)
```

4. Executing under the following environment

$ python test.py
$ dynaconf -i config.settings  write redis -v foo.bar=test` (in a separate terminal)
```

**Expected behavior**
The reason for the bug is described above. Expected result is that the dotted variables work with Fresh updates

**Comment by rochacbruno** [comment]
get_fresh doesnt support dotlookup 

Not sure if this is a bug or just undocumented behavior.

https://github.com/dynaconf/dynaconf/blob/010bd4d43fccaed66667fc1862b657c02958b4d6/dynaconf/loaders/base.py#L219 

Workaround

```python
# first take the root level key
settings.get_fresh("foo").get("bar")
```

PR #1423, opened 2026-07-16
fix: get/get_fresh with dotted keys stops working after first call

pctablet505:
## Problem

`settings.get("foo.bar", fresh=True)` (and the `get_fresh` shortcut) returns the correct value on the first call, but returns `None` on every call after that. Reported in #1187 with a Redis-backed reproduction.

## Root cause

Dotted lookups resolve one path segment at a time, recursing through `get()` for each segment. Once the top-level segment (`"foo"`) is reloaded, the recursive call that resolves the child segment (`"bar"`) against the already-fetched nested container also goes through the `fresh` unset/reload branch.

`"bar"` is not actually a top-level setting, so `unset("bar")` marks it as permanently deleted in `config.deleted`. Every later lookup of `"bar"` (including the one implicit inside `"foo.bar"`) then short-circuits to the default value before it can reach the store, since `config.deleted` is never cleared for that name.

## Fix

The fresh reload only makes sense once, on the real top-level key. Skip the unset/reload branch in `get()` when resolving a child segment against an already-loaded `parent` container (this only ever happens from `_dotted_get`'s recursion).

## Testing

Added `test_dotted_get_fresh` covering `get(..., fresh=True)`, `get_fresh`, deeper (3-level) nesting, and that fresh reload still genuinely picks up changed values from the source. Verified the new test fails without the fix and passes with it. Full non-integration suite passes.

### DIFF
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -635,8 +635,16 @@ class Settings:
             return default
 
         if (
-            fresh or config.fresh or key in config.fresh_vars
-        ) and key not in UPPER_DEFAULT_SETTINGS:
+            (fresh or config.fresh or key in config.fresh_vars)
+            and key not in UPPER_DEFAULT_SETTINGS
+            and parent is None
+        ):
+            # `parent` is only set when resolving a child segment of a
+            # dotted key against an already-fetched nested container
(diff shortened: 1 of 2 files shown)


pedro-psb (approved):
Thanks for the contribution.

PR #1426, opened 2026-07-20
fix: @json cast leaks AttributeError on non-string values

sarathfrancis90:
Calling `settings.get("SOME_MISSING_KEY", cast="@json")` blows up with `AttributeError: 'NoneType' object has no attribute 'replace'` instead of a clean dynaconf error. It happens whenever `@json` is asked to cast a non-string (a missing key resolves to `None`, but an existing dict/list/int hits it too).

The cause is in `_safe_json_parse`: its first strategy already swallows `json.loads`'s `TypeError` for non-strings, but the second strategy runs the value through a string `.replace()` and that `AttributeError` isn't caught, so it escapes before the function can raise its intended `DynaconfParseError`. `@int` and `@float` already raise `DynaconfParseError` for the same inputs, so `@json` was the odd one out.

I added `AttributeError` to that suppress block so a non-string falls through to the existing `DynaconfParseError`, and added tests covering `_safe_json_parse` directly and via `get(..., cast="@json")`.

### DIFF
--- a/dynaconf/utils/parse_conf.py
+++ b/dynaconf/utils/parse_conf.py
@@ -587,8 +587,10 @@ def _safe_json_parse(value):
         return json.loads(value)
 
     # Try extracting JSON objects with boolean/None normalization
-    # This handles single-quoted JSON and Python-style booleans
-    with suppress(json.JSONDecodeError, ValueError, TypeError):
+    # This handles single-quoted JSON and Python-style booleans.
+    # AttributeError guards against non-string input (e.g. None), where
+    # the string replacement below has no `.replace` to call.
+    with suppress(json.JSONDecodeError, ValueError, TypeError, AttributeError):
         json_objects = list(
             extract_json_objects(
                 multi_replace(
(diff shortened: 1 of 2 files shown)


dynaconf/utils/parse_conf.py
pedro-psb, line 1: The comment here is unnecessary, please remove.

tests/test_utils.py
pedro-psb, line 1: One line docstrings are enough here.

tests/test_utils.py
```diff
@@ -1399,6 +1399,31 @@ def test_safe_json_parse_complete_failure():
...
 
 
+def test_safe_json_parse_non_string_input():
+    """_safe_json_parse must raise DynaconfParseError on non-string input.
+
+    A non-string value can't be JSON parsed, so it should fail the same way
+    @int/@float do (DynaconfParseError), instead of leaking a raw
+    AttributeError from the internal string replacement step.
+    """
+    from dynaconf.utils.parse_conf import _safe_json_parse
```
pedro-psb, line 1409: Import on the module level, please.

pedro-psb (changes requested):

PR #1427, opened 2026-07-20
chore(build): switch to hatchling and finish the uv dev-environment migration

### LINKED ISSUE #1276: Use UV in the development environment
Following #1186, we should experiment and possibly addopt [UV](https://docs.astral.sh/uv/) as the 'package/project manager'.

This should not affect the end user.

Possible benefits are:
* More fast installations on CI and dev environment
* Better (dev) dependency resolution
* General utilites that uv provides on managing environemnts and python versions

addielaruee:
Finishes the remaining scope from #1276, following up on [my comment there](https://github.com/dynaconf/dynaconf/issues/1276#issuecomment-5023292885).

## What was already done (before this PR)

Dependency groups (`[dependency-groups]`: dev/ci/docs/lint/test/release) and `[tool.uv.sources]` already existed on `master` — that part of the original scope comment was already handled separately.

## What this PR adds

- **Build backend: setuptools → hatchling.** Version is still read from `dynaconf/VERSION` via `[tool.hatch.version]` (regex pattern matching the raw file content, same as the old `[tool.setuptools.dynamic]` setup). I built both the old and new backend side by side and diffed the wheel and sdist file listings to check for regressions:
  - Wheel: identical contents except a legacy setuptools-only `dist-info/top_level.txt`, which pip/`importlib.metadata` don't need.
  - Sdist: identical except `.gitignore` (harmless) and — this one's actually a fix — `tests/.env`, `conftest.py`, and `docker-compose.yml` are now correctly included. The old MANIFEST.in-based setup was silently shipping an incomplete `tests/` dir that couldn't actually run.
  - Also installed the built wheel into a scratch venv and confirmed `import dynaconf` and the `dynaconf` CLI both work.
- **Makefile: finish the "lets move all uv soon" TODO.** `install`/`ciinstall` used to `uv export` to a temp `requirements.txt` and fall back to `pip install -r` — literally flagged in a comment as temporary. Both now call `uv pip install --group` directly, no pip involved.
- **CI composite action** (`install-from-artifact`): same export-then-pip fallback, replaced with `uv pip install --system`/`--user` + `--group`. Dropped the now-unneeded "upgrade pip" steps.
- **Removed `MANIFEST.in`** — unused under hatchling; sdist inclusion is now explicit in `[tool.hatch.build.targets.sdist]`.
- **`docs/development/contributing.md`**: documents `uv venv` instead of `python3.10 -m venv venv` for environment setup, plus a link to install uv.

## Left out of scope

`smoke-test.yml`'s `pip3 install tox` — that workflow tests dynaconf against a third-party consumer repo (`galaxy_ng`), not dynaconf's own dev tooling, so it felt like a separate concern from what this issue is about. Also left the tox/mask/poethepoet task-runner question alone, as noted in my scoping comment — that seemed like a separate follow-up, not this issue.

## Testing

All run locally end to end, starting from a clean `uv venv`:

```
uv venv && source .venv/bin/activate
make install       # uv pip install -e . --group dev
make test_only      # 744 passed, 9 skipped, coverage 95.43% (required: 95%)
make mypy            # Success: no issues found in 40 source files
ruff check .          # All checks passed!

rm -rf .venv && uv venv && source .venv/bin/activate
make ciinstall        # uv pip install --upgrade -e . --group ci
make citest            # 749 passed; only failures are redis/vault integration
                        # tests needing docker services (CI provides these via
                        # service containers; I don't have them locally — unrelated
                        # to this change)
```

I wasn't able to test the Windows path of the composite action locally (no Windows environment) — the `uv pip install --user` change there is based on `uv pip install --help` documentation matching the original `pip install --user` semantics, but CI running on this PR will be the real confirmation.

Fixes #1276

### DIFF
--- a/.github/actions/install-from-artifact/action.yml
+++ b/.github/actions/install-from-artifact/action.yml
@@ -24,22 +24,11 @@ runs:
     - uses: actions/setup-python@v6
       with:
         python-version: ${{ inputs.python-version }}
-        cache: 'pip'
 
     - name: Install uv
       uses: astral-sh/setup-uv@v8.2.0
       with:
         enable-cache: true
-
-    - name: Upgrade PIP
-      if: inputs.os != 'Windows'
-      shell: bash
-      run: pip install --user --upgrade pip
-
-    - name: Upgrade PIP (Windows)
-      if: inputs.os == 'Windows'
-      shell: bash
-      run: python.exe -m pip install --user --upgrade pip
 
     - name: Download wheel
       uses: actions/download-artifact@v8
@@ -53,10 +42,9 @@ runs:
       run: |
         PACKAGE=$(ls dist/*.whl)
         if [ -n "${{ inputs.dependencies }}" ]; then
-          uv export --no-hashes --only-group ${{ inputs.dependencies }} > requirements-tmp.txt
-          pip install "$PACKAGE" -r requirements-tmp.txt
+          uv pip install --system "$PACKAGE" --group ${{ inputs.dependencies }}
         else
-          pip install "$PACKAGE"
+          uv pip install --system "$PACKAGE"
         fi
 
     - name: Install from wheel (Windows)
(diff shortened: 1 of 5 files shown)


pedro-psb:
Hey @addielaruee, I'm assesing if we really wanna change the build backend.
In any case, lint is failing.

Edit: the hatchling change is fine

pedro-psb (changes requested):
Lint is failing

pedro-psb:
Hello. Thanks for the PR, but I've been working on some related internal tasks and the whole migration is more complicated than that.

I'm superseeding this with #1449

PR #1432, opened 2026-07-24
fix: Django early validation integration code bug

pedro-psb:
## Bug
Django's `LazySettings.__getattr__` correctly raises `ImproperlyConfigured` for empty `SECRET_KEY`.
In the dynaconf-django integration code, dynaconf calls `getattr(django_settings, key, default)` for getting it's values, which triggers django lazy validations (it should'nt! it should lazily trigger only on the first app usage access, not in the integration code!).

In any case, we'd expected the early validated ImproperlyConfigured error to blow up, but Django's `ManagementUtility.execute()` silently swallows the `ImproperlyConfigured`, letting the app start without dynaconf's Wrapper.

The dynaconf integration does blow, but the app code doesn't. As a result, dynaconf doesn't reach the point of taking over the `django.settings` module, and `settings.set()` values and post_hook values are invisible through `django.conf.settings`.

## Reproducer
Django project with no `SECRET_KEY` module constant (like pulpcore 3.115+), a post_hook computing a derived setting, and a management command accessing it via `django.conf.settings`.

Two scenarios:
1. with `DJANGO_SECRET_KEY` env var: Django doesn't raise Improperly configured, because secret key is set
1. without `DJANGO_SECRET_KEY` env var: Django raise Improperly configured **only on app code**, because secret key is not set

In both cases, `django.settings` should be a dynaconf instance.

## Fix
Two changes in `django_dynaconf_v2.py`:
1. **Avoid early validation**: Step 4 reads from `django_settings._wrapped` directly instead of going through `LazySettings.__getattr__`, bypassing Django's validation during setup.
3. **Inject django's validation as a hook**: Step 5b registers an `AFTER_GET` hook that replays Django's `LazySettings.__getattr__` validations on every settings access, so checks like empty `SECRET_KEY` raising `ImproperlyConfigured` are preserved after dynaconf takes over.

### DIFF
--- a/tests_functional/issues/django_settings_set_visibility/plugin/management/commands/generate_bindings.py
+++ b/tests_functional/issues/django_settings_set_visibility/plugin/management/commands/generate_bindings.py
@@ -3,6 +3,7 @@
 Accesses computed settings via django.conf.settings, the same way
 rpm plugin code does.
 """
+
 import json
 
 from django.conf import settings
--- a/tests_functional/issues/django_settings_set_visibility/plugin/management/commands/generate_bindings.py
+++ b/tests_functional/issues/django_settings_set_visibility/plugin/management/commands/generate_bindings.py
@@ -1,0 +1,16 @@
+"""Simulates pulpcore-manager openapi --bindings --component rpm.
+
+Accesses computed settings via django.conf.settings, the same way
+rpm plugin code does.
+"""
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from dynaconf import Dynaconf
+
+class Command(BaseCommand):
+    help = "Simulates openapi bindings generation accessing computed settings"
+
+    def handle(self, *args, **options):
+        assert isinstance(settings, Dynaconf), f"{type(settings)=}"
+        root = settings.API_ROOT_NO_FRONT_SLASH
+        self.stdout.write(f"API_ROOT_NO_FRONT_SLASH={root}")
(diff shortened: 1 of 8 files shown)


tests_functional/issues/django_settings_set_visibility/plugin/management/commands/generate_bindings.py
```diff
@@ -0,0 +1,34 @@
...
+            result["secret_via_get"] = settings.get("SECRET_KEY")
+        except ImproperlyConfigured:
+            result["secret_via_get"] = "ImproperlyConfigured"
+
+        try:
+            result["secret_via_attr"] = settings.SECRET_KEY
+        except ImproperlyConfigured:
+            result["secret_via_attr"] = "ImproperlyConfigured"
+
+        self.stdout.write(json.dumps(result))
```
pedro-psb, line 34: Application code reports what it gets: the type of dynaconf instance, if was able to get secret or not, etc.

tests_functional/issues/django_settings_set_visibility/test.py
```diff
@@ -0,0 +1,58 @@
...
+assert data["api_root_no_front_slash"] == "api/v3/"
+assert data["secret_via_get"] == "test-secret-for-ci"
+assert data["secret_via_attr"] == "test-secret-for-ci"
+
+# Scenario 2: no SECRET_KEY anywhere
+data = run()
+assert data["settings_type"] == "LazySettings"
+assert data["api_root_no_front_slash"] == "api/v3/"
+assert data["secret_via_get"] == "ImproperlyConfigured"
+assert data["secret_via_attr"] == "ImproperlyConfigured"
```
pedro-psb, line 56: Assertion that the application code got what we expected.

PR #1433, opened 2026-07-25
fix: convert dict/list subclasses to data nodes

### LINKED ISSUE #1430: [bug] Regression: convert_containers skips dict subclasses, breaking lowercase_read and lazy format resolution
DataDict nodes are replaced by external box.Box during object_merge in dynaconf_hooks post hook, breaking lowercase_read and lazy format evaluation

This is preventing Robottelo's upgrade to newer dynaconf versions (still in 3.2.13).
https://github.com/SatelliteQE/robottelo/actions/runs/29763514042/job/88423786212?pr=22163#step:9:133

Environment

- dynaconf version: 3.3.2
- python-box version: 7.4.1
- Python version: 3.14 (also likely affects 3.12/3.13)

Description

When a dynaconf_hooks.py post() hook returns a dict (or Box) with dynaconf_merge=True, the object_merge function merges existing settings data into this returned dict. If the returned
container is a box.Box (external python-box), Box.__setitem__ auto-converts all nested DataDict values from the YAML config into plain box.Box objects. These plain Box objects:

1. Do not perform case-insensitive key lookup via find_the_correct_casing(), breaking lowercase_read=True
2. Do not call recursively_evaluate_lazy_format(), so @format and @jinja directives remain as raw template strings

This is a regression from the transition from DynaBox to DataDict in 3.3.x. Previously, DynaBox handled both behaviors. Now DataDict handles them, but DataDict nodes get silently replaced
by external box.Box during merge operations.

Reproduction

Minimal reproduction:

```python
# conf/repos.yaml
"""
REPOS:
RHEL_OS_REPO_HOST: http://example.com
RHEL8_OS:
    BASEOS: '@format {this[repos].rhel_os_repo_host}/rhel-8/baseos'
    APPSTREAM: '@format {this[repos].rhel_os_repo_host}/rhel-8/appstream'
YUM_1:
    URL: '@format {this[repos].rhel_os_repo_host}/fake_yum1'
"""

# conf/dynaconf_hooks.py
def post(settings):
    data = {"REPOS": {"SOME_NEW_KEY": "new_value"}}
    data["dynaconf_merge"] = True
    return data

from dynaconf import LazySettings

settings = LazySettings(
    envvar_prefix="TEST",
    core_loaders=["YAML"],
    settings_file="conf/repos.yaml",
    envless_mode=True,
    lowercase_read=True,
)

# Both of these fail:
settings.repos.rhel8_os.baseos      # BoxKeyError: "'Box' object has no attribute 'baseos'"
settings.repos.yum_1.url            # BoxKeyError: "'Box' object has no attribute 'url'"

# Uppercase access returns raw unresolved template:
settings.repos.rhel8_os.BASEOS      # '{this[repos].rhel_os_repo_host}/rhel-8/baseos'
```

Root cause trace

During execute_module_hooks("post", ...):
```
dynaconf/base.py:1254      set() -> parsed = object_merge(existing, parsed)
dynaconf/utils/__init__.py:136  object_merge() -> new[old_key] = value
box/box.py:687              Box.__setitem__ -> self.__convert_and_store(key, value)
box/box.py:587              __convert_and_store -> value = self._box_config["box_class"](value, ...)
```

object_merge(existing, parsed) iterates over existing (which contains DataDict nodes from the YAML loader) and assigns them into parsed (the dict/Box returned by the hook). When parsed is a
Box, Box.__setitem__ wraps every nested DataDict in a new box.Box, losing all DataDict behavior.

The convert_containers function in DataDict.__init__ only converts values whose __class__ is dict (identity check), not isinstance(value, dict), so Box subclasses are never converted back
to DataDict even if they pass through DataDict construction later.

Verification

```python
from dynaconf.nodes import DataDict, convert_containers
import box

# DataDict correctly converts plain dicts
data = DataDict({"nested": {"key": "value"}}, core=None)
type(data["nested"])  # <class 'dynaconf.nodes.DataDict'> ✓

# But DataDict does NOT convert Box objects (Box.__class__ is not dict)
data = DataDict({"nested": box.Box({"key": "value"})}, core=None)
type(data["nested"])  # <class 'box.box.Box'> ✗
```

Impact

This breaks any project that:
1. Uses dynaconf_hooks.py with a post() hook that triggers dynaconf_merge
2. Relies on lowercase_read=True for case-insensitive access to nested settings
3. Uses @format or @jinja directives in nested dict values within YAML configs

All three behaviors silently stop working — no error is raised during loading, only when the affected settings are accessed.

Possible fix

convert_containers in dynaconf/nodes.py should use isinstance rather than identity checks, or should also handle Box subclasses:

```python
def convert_containers(data, iter, core):
    for key, value in iter:
        if isinstance(value, dict) and not isinstance(value, DataDict):
            data[key] = DataDict(value, core=core)
        if isinstance(value, list) and not isinstance(value, DataList):
            data[key] = DataList(value, core=core)
```

Alternatively (or additionally), object_merge could avoid assigning into Box containers by converting the merge target to a plain dict first.

Sanjays2402:
Closes #1430

`convert_containers` used an identity check (`value.__class__ is dict`), so subclasses of `dict`/`list` -- `OrderedDict`, external `box.Box`, or any user subclass -- were never converted to `DataDict`/`DataList`. Those plain containers do not do case-insensitive lookup and do not evaluate lazy values, so nested settings reaching them lose `lowercase_read` and keep `@format`/`@jinja` directives as raw strings.

Switched to `isinstance()`, skipping values that are already data nodes.

Regression test `test_nested_dict_and_list_subclasses_are_converted` fails on the original check and passes with the fix.

### DIFF
--- a/dynaconf/nodes.py
+++ b/dynaconf/nodes.py
@@ -702,9 +702,9 @@ def get_core(node, raises=True) -> DynaconfCore:
 
 def convert_containers(data: dict | list | DataNode, iter, core):
     for key, value in iter:
-        if value.__class__ is dict:
+        if isinstance(value, dict) and not isinstance(value, DataDict):
             data[key] = DataDict(value, core=core)
-        if value.__class__ is list:
+        elif isinstance(value, list) and not isinstance(value, DataList):
             data[key] = DataList(value, core=core)
 
 
--- a/tests/test_nodes.py
+++ b/tests/test_nodes.py
@@ -288,6 +288,28 @@ def test_nested_structures():
     assert isinstance(di["b"]["c"], DataList)
 
 
+def test_nested_dict_and_list_subclasses_are_converted():
+    """Subclasses of dict/list must become data-nodes too.
+
+    Regression for #1430: an identity check on ``__class__`` skipped
+    dict/list subclasses (e.g. ``OrderedDict``, ``box.Box``), so nested
+    values kept plain-container behavior and lost case-insensitive
+    reading and lazy-format evaluation.
+    """
+
+    class MyDict(dict):
+        pass
+
+    class MyList(list):
+        pass
+
+    di = DataDict({"a": MyDict({"x": 1}), "b": MyList([{"y": 2}])})
+
+    assert isinstance(di["a"], DataDict)
+    assert isinstance(di["b"], DataList)
+    assert isinstance(di["b"][0], DataDict)
+
+
 def test_method_preservation():
     di = DataDict()
     di["a"] = {"x": 1}


pedro-psb:
Already fixed in https://github.com/dynaconf/dynaconf/pull/1436

PR #1440, opened 2026-07-31
fix: add __deepcopy__ to DataDict to match DataList

### LINKED ISSUE #1439: [bug] DataDict missing __deepcopy__ — copy.deepcopy() crashes with "cannot pickle '_thread.RLock' object"
## Bug Description

`DataDict` does not implement `__deepcopy__`. Calling `copy.deepcopy()` on a `DataDict` recurses into `__meta__.core` → `DynaconfCore` → `Settings` and crashes when the Settings graph holds non-picklable objects (thread locks, DI containers, logging handlers, etc.).

`DataList` already handles this correctly with its own `__deepcopy__` (nodes.py), but the equivalent is missing on `DataDict`.

## How to reproduce

**settings.yaml:**
```yaml
SERVER:
  host: localhost
  port: 8080
```

**reproduce.py:**
```python
import copy
import threading
from dynaconf import Dynaconf

settings = Dynaconf(settings_files=["settings.yaml"])
settings._lock = threading.RLock()  # common in real apps (DI, caches, etc.)

server = settings.SERVER  # returns a DataDict

copy.deepcopy(server)  # TypeError: cannot pickle '_thread.RLock' object
```

**Output (dynaconf 3.3.4, Python 3.14):**
```
TypeError: cannot pickle '_thread.RLock' object
```

## Root Cause

`DataDict` defines `__copy__` but not `__deepcopy__`:

```python
class DataDict(dict):
    def __init__(self, *args, **kwargs):
        core = kwargs.pop("box_settings", kwargs.pop("core", None))
        self.__meta__ = NodeMetadata(core=core)
        ...

    def __copy__(self):
        return self.copy()

    # __deepcopy__ is MISSING
```

Python's default `deepcopy` recurses into `__meta__` → `NodeMetadata.core` → `DynaconfCore` → `Settings` → non-picklable objects → crash.

`DataList` already solves this by propagating `core` by reference:

```python
class DataList(list):
    ...
    def __deepcopy__(self, memo=None):
        out = self.__class__(core=self.__meta__.core)
        memo = memo or {}
        memo[id(self)] = out
        for k in self:
            out.append(copy.deepcopy(k, memo=memo))
        return out
```

## Suggested Fix

Add the equivalent `__deepcopy__` to `DataDict`:

```python
def __deepcopy__(self, memo=None):
    memo = memo or {}
    out = self.__class__(box_settings=self.__meta__.core)
    memo[id(self)] = out
    for k, v in dict.items(self):
        out[copy.deepcopy(k, memo)] = copy.deepcopy(v, memo)
    return out
```

## Impact

Affects any dynaconf >= 3.3.0 user who calls `copy.deepcopy()` on config values — a common pattern for Kafka message building, test isolation, multiprocessing, or any code that assumes `dict` subclasses support the standard copy protocol.

## Environment

- **dynaconf:** 3.3.0 – 3.3.4 (all affected)
- **Python:** 3.12, 3.14 (tested both)
- **Introduced in:** #1325

## Workaround

```python
from dynaconf.utils import to_dict

copy.deepcopy(to_dict(server))  # strips DataDict wrappers first
```

digitronik:
Fix for https://github.com/dynaconf/dynaconf/issues/1439
closes: #1439

### DIFF
--- a/dynaconf/nodes.py
+++ b/dynaconf/nodes.py
@@ -89,6 +89,15 @@ class DataDict(dict):
     def __copy__(self):
         return self.copy()
 
+    def __deepcopy__(self, memo=None):
+        memo = memo or {}
+        out = self.__class__(core=self.__meta__.core)
+        out.__meta__.data_env = self.__meta__.data_env
+        memo[id(self)] = out
+        for k, v in dict.items(self):
+            out[copy.deepcopy(k, memo=memo)] = copy.deepcopy(v, memo=memo)
+        return out
+
     def __getitem__(self, item):
         try:
             result = super().__getitem__(item)
--- a/tests/test_nodes.py
+++ b/tests/test_nodes.py
@@ -252,6 +252,38 @@ class TestDataDict:
         with pytest.raises(AccessError):
             _ = di.nonexistent_key
 
+    def test_deepcopy(self):
+        core = DynaconfCore("test")
+        di = DataDict({"a": 1, "b": {"c": 2}}, core=core)
+
+        di_copy = copy.deepcopy(di)
+
+        assert di_copy == di
+        assert di_copy is not di
+        assert isinstance(di_copy, DataDict)
+        assert isinstance(di_copy["b"], DataDict)
+        assert get_core(di_copy) is core
+
+    def test_deepcopy_does_not_traverse_core(self):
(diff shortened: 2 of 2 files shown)


pedro-psb (approved):
Thanks for the contribution!

PR #1446, opened 2026-08-04
fix: skip non-str keys when resolving key casing

shuvamk:
`find_the_correct_casing` returns on the first non-str key in the data, so a non-str key disables case-insensitive matching for the keys after it.

`settings.yaml`:

```yaml
default:
  RETRY:
    1: fast
    2: slow
    TIMEOUT: 30
```

`settings.local.yaml`:

```yaml
default:
  RETRY:
    timeout: 60
```

```python
from dynaconf import Dynaconf

s = Dynaconf(settings_files=["settings.yaml", "settings.local.yaml"],
             environments=True, merge_enabled=True)
print("RETRY =", dict(s.RETRY))
print("RETRY.TIMEOUT =", s.RETRY.TIMEOUT)
```

On master:

```
RETRY = {1: 'fast', 2: 'slow', 'TIMEOUT': 30, 'timeout': 60}
RETRY.TIMEOUT = 30
```

The override is dropped and a duplicate key appears; the same two files with `A:`/`B:` instead of `1:`/`2:` give `{'A': 'fast', 'B': 'slow', 'TIMEOUT': 60}`. The same lookup also backs `__getitem__`, `__getattr__`, `__contains__` and `get`.

The guard is there to avoid `k.lower()` on a non-str key, and `continue` does that without aborting the scan. A dict whose keys are all non-str still resolves to `None`, now at the end of the loop.

Two tests in `tests/test_utils.py`; I verified both fail without the one-line change. Unit and functional suites pass, mypy and pre-commit clean.

### DIFF
--- a/dynaconf/utils/__init__.py
+++ b/dynaconf/utils/__init__.py
@@ -617,7 +617,7 @@ def find_the_correct_casing(
 ) -> str | None:
     """Given a key, find the proper casing in data.
 
-    Return 'None' for non-str key types.
+    Non-str keys in data are skipped.
 
     Arguments:
         key {str} -- A key to be searched in data
@@ -630,7 +630,7 @@ def find_the_correct_casing(
         return key
     for k in data_keys:
         if not isinstance(k, str):
-            return None
+            continue
         if k.lower() == key.lower():
             return k
         if k.replace(" ", "_").lower() == key.lower():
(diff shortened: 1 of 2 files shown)


pedro-psb (approved):
Nice catch, thanks for the contribution.

PR #1417, opened 2026-06-30
chore(ci): Consolidate docs-publishing workflow for rolling-release

pedro-psb:
* Added standalone ruff config for release script, so it doesn't conflict with the codebase style (e.g, for backport branches)
* Backport branches don't need to build docs, as they are only published via rolling releases from master
* Docs publication was extracted to a standalone workflow, so it can be re-triggered manually if required
* Removed the config for the Github App based netlify deployment, as we don't use the docs preview anymore and the prod deployment is unintuitive to configure and trigger (very vendor-lock in workflow).
    * Actually, after the PR merge, it'll require disabling the netlify App in the repository settings as well
    * This was done in favour of direct netlicy cli usage

PR #1447, opened 2026-08-13
fix: run CI workflow on release branches, not just master

pedro-psb:
## Summary
- The repo's "Base Protection" ruleset requires the `checkpoint` status check on any ref matching `refs/heads/*.*` (in addition to the default branch) — this pattern matches release branches like `3.2`, `3.1`, etc.
- `.github/workflows/main.yml` (which produces the `checkpoint` check) only triggers on `push`/`pull_request` events targeting `master`, so PRs opened against release branches (e.g. backports like #1445) never run CI at all and never satisfy the required check.
- This fixes the trigger to also match `*.*` branches, aligning it with the ruleset's own pattern.

## Test plan
- [ ] Confirm CI (`checkpoint`) runs on a PR targeting a release branch (e.g. `3.2`) after this merges and syncs via `ci-update.yml`
- [ ] Confirm CI still runs normally on PRs targeting `master`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pedro-psb:
Closing — this fix already exists on master (merged via #1417) and is already staged for 3.2 in the pending sync PR #1405, which just needs to be merged.

PR #1448, opened 2026-08-14
fix: allow lazy self-reference to read the previous value

### LINKED ISSUE #1425: [bug] DynaconfFormatError (circular reference) when applying @jinja to alter the same value
**Describe the bug**
I apply the following code to make sure that a value is an integer even if defined as string:
`MAIN__STARTYEAR = "@int @jinja {{this.main.startyear | int}}"`
This is part of an toml-file defined via the includes-parameter of `Dynaconf`

Before version 3.3.0 this went well, now it raises a DynaconfFormatError (Circular reference detected in lazy formatting. A value is referencing itself directly or indirectly.) in nodes.py:734

**To Reproduce**

1. Having the following folder structure

<details>
<summary> Project structure </summary>

```bash
# /dynaconf_circular/
```
</details>

2. Having the following config files:

<details>
<summary> Config files </summary>

**/dynaconf_circular/settings.toml**
```toml
[main]
startyear = "2015"
```
and

**/dynaconf_circular/config_post.toml**
```bash
MAIN__STARTYEAR = "@int @jinja {{this.main.startyear | int}}"
```
</details>

3. Having the following app code:

<details>
<summary> Code </summary>

**/dynaconf_circular/config.py**
```python
from dynaconf import Dynaconf

if __name__ == '__main__':
    settings_dyna = Dynaconf(
        merge_enabled=True,
        environments=False,
        settings_files=["settings.toml"],
        loaders=[
            "dynaconf.loaders.toml_loader",
            ],
        includes=["config_post.toml"],
    )
    print(settings_dyna.main.startyear)
```

</details>

4. Executing under the following environment

<details>
<summary> Execution </summary>

```bash

$ python config.py
```

</details>

**Expected behavior**
Printing "2015"

 - OS: Windows 11
 - Dynaconf Version 3.3.0

**Additional context**
For the DynaconfFormatError, it would be good to include more context (i.e. the value that is circularly referenced) in the error message!

DSeaStar:
## Summary

A lazy `@jinja` / `@format` value that overwrites an existing key and then reads that same key (for example `FOO="@int @jinja {{this.FOO | int}}"` after `FOO="2015"`) was reported as a circular reference and raised `DynaconfFormatError`.

That pattern is used to recast or transform a value that was already loaded (issue #1425). The 3.3 circular-reference guard treated the self-read as a cycle because the previous value had already been replaced in the store.

## Change

- When `set()` / dotted set stores a `Lazy` over an existing key, keep the replaced (unevaluated) value on `Lazy.previous`.
- If evaluation re-enters the same `Lazy` object, return that previous value instead of raising.
- True cycles (`A` → `B` → `A`, or a self-ref with no previous value) still raise. The error now includes the lazy value.

## Tests

- Top-level `set()` self-reference with `@int @jinja` and `@format`
- Nested / include reproduction from #1425
- Existing A→B→A circular test still fails as expected

Fixes #1425

### DIFF
--- a/dynaconf/base.py
+++ b/dynaconf/base.py
@@ -35,6 +35,7 @@ from dynaconf.nodes import recursively_evaluate_lazy_format
 from dynaconf.strategies.filtering import PrefixFilter
 from dynaconf.utils import BANNER
 from dynaconf.utils import ensure_a_list
+from dynaconf.utils import find_the_correct_casing
 from dynaconf.utils import ensure_upperfied_list
 from dynaconf.utils import ListMergeOptions
 from dynaconf.utils import missing
@@ -1010,6 +1011,22 @@ class Settings:
         for key in keys:
             self.unset(key, force=force)
 
+    def _get_raw_leaf(self, split_keys):
+        """Return the unevaluated value at a dotted path, or `empty`."""
+        node = self.store
+        try:
+            for k in split_keys:
+                if isinstance(node, DataDict):
+                    node = node.get(k, bypass_eval=True)
+                elif isinstance(node, dict):
+                    n_item = find_the_correct_casing(k, tuple(node.keys())) or k
+                    node = node[n_item]
+                else:
+                    return empty
+            return node
+        except (AttributeError, KeyError, TypeError):
+            return empty
+
     def _dotted_set(
         self,
         dotted_key: str,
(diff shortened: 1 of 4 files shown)


pedro-psb:
@DSeaStar  Lint is failing

PR #1449, opened 2026-08-14
Re-structure Makefile, uv usage and  CI/dev consistency

### LINKED ISSUE #1276: Use UV in the development environment

### LINKED ISSUE #1401: Investigate flaky `tests_functional/legacy/django_pytest`
I've hit this myself on a local setup and but I dindt understand the root cause.
My suspicious is that it might have to do with django version not being pinned and the use of uv.

If that's the case, we should design the tests so this is reproducible and understand how uv can be used.
More broadly, we can improve how we "support" major django versions and how that reflect on the tests.

Recently a user also hit the error when running `uv run make test_all`.
The error din't show in CI.

It is a sqlite table: `no such table: auth_user`

pedro-psb:
This changes the makefile in significant ways, dropping some commands and merging others.
Here is how different the help looks.

The main motivation was to standardize CI with local development when using UV.
Besides that, I've added some conveniences for the development workflow and changed some some names with my optionated opinions hehe

Some highlights:

* In the dev machine, `make install` manages a `.venv`, which can be easily destroyed, replaced and configured with different python and extra options. E.g, we can very easily do `make build && make install WHELL=dist/*.whl` to install from the built wheel, as CI does,
* Commands such as `test` will use that, so it's easy to test different envs manually
* Added [actionlint](https://github.com/rhysd/actionlint) pre-commit hook, just because
* Fxed functional tests that used ephemeral `pip install/uninstall` (TODO: rethink venv isolation in the functional tests)
* Fixed integrations tests (redis and vault). They where silently failing because of some some conflict between the script `docker run` and the worklfow `services:` section.
* Re-structure test jobs grouping into test types (unit, functional, integration, and paralellize.
    * Each job/test-type has a matrix of python versions and OS (os-specific tasks are skipped/selected explicitly)
    * parallelize jobs:
        * test-install + test-unit
        * test-functional + test-integration

It also fixed a bug in CI where we were not really running against different python versions.

See the changes in the Make help:

```
=== make help: BEFORE (commit 85e501a) ===
all
bench
bump-minor
check-releases
ciinstall
citest
clean
coverage-report
dist
docs
install
minify_vendor
mypy
publish
run-pre-commit
run-tox
setup-pre-commit
source_vendor
test
test_all
test_functional
test_integration
test_only
test_redis
test_vault
watch
watch_coverage
watch_django
watch_test

=== make help: AFTER (current working tree) ===
General
    all               Clean, install dev deps, lint, and run the full test suite
    clean             Remove build artifacts, caches, and generated files
    docs              Serve the documentation site locally with live-reload
    docs-build        Build the documentation site
    lint              Lint and format the code
Setup
    info              Show info of local venv (.venv) | [VERBOSE=1] full pip list
    install           Create .venv and install project | [WHEEL=<path>] [GROUP=<group>] [PYTHON=<version>] [ARGS=<uv-args>]
    tips              Show quick usage tips
Tests
    bench             Run performance benchmarks
    test              Run the unit test suite with coverage | [k=<expr>] run tests matching <expr>
    test-all          Run every test suite and generate an HTML coverage report
    test-functional   Run the functional test suite | [k=<expr>] run tests matching <expr>
    test-integration  Run integration-marked tests only | [k=<expr>] run tests matching <expr>
    test-redis        Run functional tests against a local Redis
    test-vault        Run functional tests against a local Vault
Build & Release
    build             Build the sdist and wheel
    bump-minor        Bump X.Y.Z.dev to X.(Y+1).0.dev | [YES=1] create the bump commit (DEFAULT=dry-run)
    check-releases    Check for pending releases | Release via https://github.com/dynaconf/dynaconf/actions)
    release-notes     Preview release notes
```

Closes #1401 
Closes #1276

### DIFF
--- a/scripts/minify.sh
+++ b/scripts/minify.sh
@@ -2,26 +2,6 @@
 
 set -euo pipefail
 
-# Ensure we are using a python lowerbound version
-PYTHON_LOWERBOUND=$(python -c \
-   'import toml; \
-   data=toml.load("pyproject.toml"); \
-   version=data["project"]["requires-python"].split(",")[0].split("=")[-1]; \
-   print(version)' \
-)
-
-PYTHON_ERR_MSG="$(cat <<EOF
-ERROR: Must use python ${PYTHON_LOWERBOUND}
-Minification creates code compatible with the current python version in use.
-For greater compatibility, the build should be done with the lower python we support.\n
-EOF
-)"
-
-if ! python --version | grep $PYTHON_LOWERBOUND; then
-   printf "$PYTHON_ERR_MSG"
-   exit 1
-fi
-
 # Ensure vendor is source and cleanup vendor_src backup folder
 ls dynaconf/vendor/source && rm -rf dynaconf/vendor_src
 
@@ -47,7 +27,7 @@ do
    touch dynaconf/vendor/$direc/__init__.py
 
    # for each .py file in the vendor_srx directory
-   # NOTE: vendor_src is created when make minify_vendor is run
+   # NOTE: vendor_src is created when make build runs ./scripts/minify.sh
    for eachfile in `ls dynaconf/vendor_src/$direc/*.py`
    do
       # minify each file pasting resilts to the dynaconf/vendor directory
--- a/tests_functional/legacy/django_pytest_pure/Makefile
+++ b/tests_functional/legacy/django_pytest_pure/Makefile
@@ -1,6 +1,6 @@
 .PHONY: test
 
 test:
-	pip install pytest-django
+	uv pip install pytest-django
 	DJANGO_SETTINGS_MODULE=project.settings DJANGO_ENVIRONMENT=default pytest
-	pip uninstall -y pytest-django
+	uv pip uninstall pytest-django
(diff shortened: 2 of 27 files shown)


scripts/minify.sh
pedro-psb, line 1: Removing the check here, because `make build` already enforces we use the lowerbound python for doing the build (that matters, because minify may generate code incompatible with our declared python lowerbound if we build with upperbound python).

tests_functional/legacy/django_pytest_pure/Makefile
pedro-psb, line 1: Ideally each of these functional tests should have their dependency (except from dynaconf) declared explicitly, and use their own virtualenv for running! But it's fine for now, one step at a time.

tests_functional/test_redis.sh
pedro-psb, line 1: This (and `test_vault*.sh`) where silently failing because of workflow services and this docker run conflict (something like port is already being used). But they passed because there was not `set -euo pipefail`

rochacbruno (approved):

PR #1451, opened 2026-08-17
chore(python): add support for python 3.15

kytta:
Python 3.15.0 candidate 1 is out. "candidate" means that there will be no more ABI changes, and the API over all is now stable. It is now pretty much safe to test against and to specify support for it.

This PR adds Python 3.15 support to Dynaconf:
- version and classifier in `pyproject.toml`
- `3.15-dev` (to be replaced with `3.15` in October) in GitHub Actions

I have ran the tests locally, and all of them pass.

This PR _keeps_ the upper bound on `requires-python`, but as stated in #1452, I highly recommend you remove it :)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,7 +54,7 @@ jobs:
     with:
       dist-artifact: ${{ needs.build.outputs.dist-artifact }}
       python_minimal: '["3.10", "3.14"]'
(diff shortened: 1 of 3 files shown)


pedro-psb (approved):
Thanks for the contribution!
We'll consider removing the upperbound.

PR #1455, opened 2026-08-23
feat(vault): pin a KV v2 secret version (VAULT_SECRET_VERSION_FOR_DYNACONF)

ahridin:
## Summary

The KV v2 engine keeps a version history for every secret, but the vault loader always reads the latest one: `read_secret_version()` is called without its `version` argument. There's currently no way to ask for anything but the current version.

This adds **`VAULT_SECRET_VERSION_FOR_DYNACONF`** (default `None`), passed through to `hvac`'s `read_secret_version(version=...)`.

- `None` keeps today's behaviour exactly — `version=None` is hvac's own "latest" default, so unpinned users see no change.
- A path that has no such version is skipped, exactly like a path that does not exist (hvac maps Vault's 404 to `InvalidPath`, which the loader already handles).
- A pin combined with `VAULT_KV_VERSION_FOR_DYNACONF=1` raises `ValueError` **before any network call**: KV v1 keeps no versions, and silently reading the latest while claiming a pin would be worse than refusing. This mirrors the existing precedent of `get_client()` raising loudly on misconfiguration regardless of `silent`.

## Motivation / use case

Sometimes a specific version of a secret is needed rather than whatever is current — for example, `API_KEY` was `abc123` at version 3 and got rotated to `xyz789` at version 4, but something still needs to read the value at version 3. Vault's own KV v2 API, the `vault kv get -version=N` CLI, and hvac all support reading a specific version directly; the loader was the only link in the chain that could not express it.

## Changes

- `dynaconf/loaders/vault_loader.py` — read the setting, validate against KV v1 (before `get_client`, so it fails fast and offline), pass `version=` to `read_secret_version`.
- `dynaconf/default_settings.py` — `VAULT_SECRET_VERSION_FOR_DYNACONF = None`.
- `docs/secrets.md` — "Pinning a secret version (KV v2)" section, including the note that the pin applies to every path the loader walks, so it is most useful with a single secret path.
- `tests/test_vault.py` — three integration tests: pinned read returns the first revision after a second write; a missing pinned version is skipped like a missing path; a pin on KV v1 raises.
- `tests_functional/legacy/vault/vault_example.py` — end-to-end: second revision written, version 1 read back through the pin, KV v1 refusal asserted.

## Notes for review

- The same pinned version is requested for every path the loader iterates. Documented as such; per-path pins would need a much bigger interface and the single-path case is the one that motivates the feature.
- `raise_on_deleted_version=True` is left as-is; a pin to a deleted version raises, which seems right for something asked for by exact number.

## Testing

Ran the real integration suite against a live Vault container
(`tests/test_vault.py -m integration`, the same fixture `make test-integration`
uses): 11/12 pass, including the 3 new tests. The one remaining failure,
`test_load_vault_not_configured`, is unrelated and pre-existing — confirmed
by running it in isolation against unmodified upstream `master`, where it
fails identically (it implicitly relies on Vault already being warm from an
earlier fixture, only true in a full multi-file run).

The 3 new tests needed a fix once actually run as part of the full file
rather than in isolation: they originally shared the suite's default Vault
path and left `VAULT_KV_VERSION_FOR_DYNACONF` set in `os.environ` between
tests, which (env vars outrank constructor kwargs in dynaconf's own
precedence) silently interfered with a later test. Isolated to a dedicated
path and cleaned up.


ahridin:
Friendly nudge — this PR has been open a while with no CI run yet (likely waiting on a maintainer to approve the workflow run for an external contributor) and no review. Happy to address any feedback. Thanks for taking a look!

ahridin:
@pedro-psb — could you take a look when you have a moment? The CI run is gated as `action_required` (fork PR from a first-time contributor), so it's never actually executed — needs an approval click under Actions → this run → "Approve and run workflow". Happy to make any changes needed for review. Thanks!

ahridin:
**Current test report** (full `tests/test_vault.py -m integration` against a live `hashicorp/vault:1.21` container, head `f522abb`):

```
test_write_vault_not_configured                       PASSED
test_write_vault_without_data                         PASSED
test_list_envs_in_vault                               PASSED
test_write_to_vault                                   PASSED
test_load_from_vault_with_key                         PASSED
test_write_and_load_from_vault_without_key            PASSED
test_read_from_vault_kv2_with_different_environments  PASSED
test_read_pinned_secret_version_kv2                   PASSED  <- new
test_pinned_version_missing_is_skipped                PASSED  <- new
test_pinned_version_requires_kv2                      PASSED  <- new
test_vault_has_proper_source_metadata                 PASSED
test_load_vault_not_configured                        FAILED  (pre-existing)
```

11/12 — the single failure is unrelated and reproduces identically on unmodified `master` run in isolation (it implicitly depends on Vault being warm from an earlier fixture, which only holds in a full multi-file `make test-integration` run).

